PR #25, opened 2026-06-05
Support GCS range downloads

### LINKED ISSUE #19: [BUG] Cloud Storage media downloads ignore Range header
## Service

Cloud Storage (GCS)

## GCP API / Method

`objects.get` media download (`GET /storage/v1/b/{bucket}/o/{object}?alt=media`) with the HTTP `Range` header

## Expected behavior

`objects.get` with `alt=media` and a satisfiable HTTP `Range` header should return the requested byte range only. For an object containing `0123456789abcdef`, `Range: bytes=4-7` should return `4567` and the response should indicate partial content rather than returning the full object.

Verified against the Cloud Storage API: the request returned HTTP 206, `Content-Range: bytes 4-7/16`, and body `4567`.

## Actual behavior

The emulator returns the full object with HTTP 200:

```text
HTTP/1.1 200 OK
content-length: 16
Content-Type: application/octet-stream

0123456789abcdef
```

## Reproduction

```bash
docker run --rm -p 4588:4588 floci/floci-gcp:0.2.0

curl -X POST 'http://127.0.0.1:4588/storage/v1/b?project=floci-local' \
  -d '{"name":"floci-repro"}'

printf '0123456789abcdef' > /tmp/range.txt
curl -X POST 'http://127.0.0.1:4588/upload/storage/v1/b/floci-repro/o?uploadType=media&name=range.txt' \
  -H 'Content-Type: application/octet-stream' \
  --data-binary @/tmp/range.txt

curl -D - -H 'Range: bytes=4-7' \
  'http://127.0.0.1:4588/storage/v1/b/floci-repro/o/range.txt?alt=media'
```

## Environment

- floci-gcp version / image tag: `floci/floci-gcp:0.2.0`
- GCP SDK version (if applicable): Not applicable
- How you're running floci-gcp (Docker / native / `mvn quarkus:dev`): Docker

electrum:
## Summary

Support HTTP `Range` requests for Cloud Storage media download routes.

Closes #19

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

Incorrect behavior: media downloads returned the full object with HTTP 200 even when the request included a satisfiable `Range` header, and an unsatisfiable start offset returned the wrong error class for SDK callers.

The download routes now return HTTP 206 with `Content-Range` and the requested bytes. Unsatisfiable ranges now return HTTP 416 with `OUT_OF_RANGE`.

Checked against the Cloud Storage API: `Range: bytes=4-7` for an object containing `0123456789abcdef` returned HTTP 206, `Content-Range: bytes 4-7/16`, and body `4567`. Covered locally by `GcsRangeDownloadRawTest` for JSON media, `/download/storage/v1`, XML-style object downloads, and a past-end range returning HTTP 416.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/io/floci/gcp/test/GcsRangeDownloadRawTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/floci/gcp/test/GcsRangeDownloadRawTest.java
@@ -1,0 +1,83 @@
+package io.floci.gcp.test;
+
+import com.google.cloud.storage.BlobId;
+import com.google.cloud.storage.BlobInfo;
+import com.google.cloud.storage.BucketInfo;
+import com.google.cloud.storage.Storage;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.nio.charset.StandardCharsets;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class GcsRangeDownloadRawTest {
+
+    private static final String BUCKET = TestFixtures.uniqueName("range-bucket");
+    private static final String OBJECT = "range.txt";
+    private static final HttpClient CLIENT = HttpClient.newHttpClient();
+
+    private static Storage storage;
+
+    @BeforeAll
+    static void setUp() {
+        storage = TestFixtures.storageClient();
+        storage.create(BucketInfo.of(BUCKET));
+        storage.create(BlobInfo.newBuilder(BlobId.of(BUCKET, OBJECT)).build(),
+                "0123456789abcdef".getBytes(StandardCharsets.UTF_8));
+    }
+
+    @AfterAll
+    static void tearDown() throws Exception {
+        if (storage instanceof AutoCloseable closeable) {
+            closeable.close();
+        }
+    }
+
+    @Test
+    void mediaDownloadHonorsRangeHeader() throws Exception {
+        assertRangeResponse(URI.create(TestFixtures.endpoint()
+                + "/storage/v1/b/" + BUCKET + "/o/" + OBJECT + "?alt=media"));
+    }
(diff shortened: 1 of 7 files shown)


hectorvent:
Thanks for tackling this! The refactor into `GcsMediaResponses` is really clean, and routing all three download paths through one helper is exactly the right call. The happy path looks solid: closed ranges, open ended ranges, suffix ranges, and end clamping all behave correctly, and `bytes 4-7/16` matches real GCS.

One point I'd treat as blocking, plus a couple of optional notes.

**Unsatisfiable range should return 416, not 400 (SDK compatibility)**

When the start is at or past the end of the object (for example seeking to `length` on a 16 byte object), real GCS returns `416 Range Not Satisfiable`. This is not just spec pedantry: the Java SDK read channel specifically depends on it. `BaseStorageReadChannel` and `ApiaryUnbufferedReadableByteChannel` both special case a 416 and translate it into a clean EOF (`read` returns -1), and there are integration tests covering exactly this (`responseWith416ReturnsZeroAndLeavesTheChannelOpen` and friends). If we return 400 instead, the SDK raises an `IOException` rather than hitting EOF, so a real client reading to the end of an object would break.

One heads up for the fix: `GcpException` does not have a 416 helper today, so this would need a small new factory (something like `outOfRange(...)` mapping to 416 / `OUT_OF_RANGE`) rather than reusing `invalidArgument`.

**Optional, lower confidence: malformed and multi range headers**

Per RFC 7233, a Range header the server cannot parse is usually ignored and the full object is returned with 200, and GCS does not support multipart ranges. Currently those cases return 400. I could not confirm GCS's exact behavior here from the SDK source, so treat this as a suggestion rather than a requirement, but falling back to a full 200 on an unparseable header would be a safe, client friendly default.

**Tests**

Coverage hits the `bytes=4-7` happy path across all three routes, which is great. Since the parser has a few branches, could we add an out of range start case asserting 416? That one matters most given the point above. Open ended and suffix cases would be nice bonuses. Also worth confirming `./mvnw test` passes locally, since the checklist box is still unchecked.

Overall really nice work, this is close. The main thing is the 416 status for reads past the end, since the SDK relies on it.

PR #27, opened 2026-06-05
Keep GCS resumable uploads open

### LINKED ISSUE #23: [BUG] Cloud Storage resumable upload finalizes after first non-final chunk
## Service

Cloud Storage (GCS)

## GCP API / Method

`objects.insert` resumable upload (`uploadType=resumable`)

## Expected behavior

For `objects.insert` resumable uploads, a non-final chunk with `Content-Range: bytes 0-262143/*` should keep the upload session open and return a resumable-upload progress response. The object should only be finalized after a request that supplies the final byte range and total object size.

Verified against the Cloud Storage API: the non-final chunk returned HTTP 308 with `Range: bytes=0-262143`, and the final chunk returned HTTP 200 with object size `263168`.

## Actual behavior

The first non-final chunk finalizes the object and returns HTTP 200 with object metadata:

```text
HTTP/1.1 200 OK
...
"size":"16777216"
```

A subsequent chunk to the same upload URL returns HTTP 404:

```json
{
  "error": {
    "code": 404,
    "message": "Resumable upload not found: <upload_id>",
    "status": "NOT_FOUND"
  }
}
```

## Reproduction

```bash
docker run --rm -p 4588:4588 floci/floci-gcp:0.2.0

curl -X POST 'http://127.0.0.1:4588/storage/v1/b?project=floci-local' \
  -d '{"name":"floci-repro"}'

python3 - <<'PY'
from pathlib import Path
Path('/tmp/chunk1.bin').write_bytes(b'a' * (16 * 1024 * 1024))
Path('/tmp/chunk2.bin').write_bytes(b'b' * (1024 * 1024))
PY

curl -D /tmp/resumable-init.headers -o /tmp/resumable-init.body \
  -X POST 'http://127.0.0.1:4588/upload/storage/v1/b/floci-repro/o?uploadType=resumable&name=resumable.bin' \
  -H 'X-Upload-Content-Type: application/octet-stream' \
  -H 'X-Upload-Content-Length: 17825792' \
  -d '{}'

upload_url=$(grep -i '^Location:' /tmp/resumable-init.headers | sed -e 's/^[^:]*:[[:space:]]*//' -e $'s/\r$//')

curl -D - -X PUT "$upload_url" \
  -H 'Content-Type: application/octet-stream' \
  -H 'Content-Range: bytes 0-16777215/*' \
  --data-binary @/tmp/chunk1.bin

curl -D - -X PUT "$upload_url" \
  -H 'Content-Type: application/octet-stream' \
  -H 'Content-Range: bytes 16777216-17825791/17825792' \
  --data-binary @/tmp/chunk2.bin
```

## Environment

- floci-gcp version / image tag: `floci/floci-gcp:0.2.0`
- GCP SDK version (if applicable): Not applicable
- How you're running floci-gcp (Docker / native / `mvn quarkus:dev`): Docker

electrum:
## Summary

Keep Cloud Storage resumable upload sessions open until the final chunk.

Closes #23

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

Incorrect behavior: a non-final resumable upload chunk finalized the object and removed the upload session. The Node.js client default upload path also sends `Content-Range: bytes 0-*/*`, which must finalize the upload instead of failing header parsing.

Non-final chunks now return HTTP 308 with progress, and the object is finalized only after a final chunk. Unknown-size final chunks are accepted, and zero-byte status checks with `Content-Range: bytes */...` now return HTTP 308 with a `Range` header only after bytes have been persisted.

Checked against the Cloud Storage API: `Content-Range: bytes 0-262143/*` returned HTTP 308 with `Range: bytes=0-262143`, and the final chunk returned HTTP 200 with object size `263168`. Covered locally by `GcsResumableUploadRawTest`, including status checks and `Content-Range: bytes 0-*/*`. The Node.js Cloud Storage GCS suite also passes locally against the packaged emulator.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/io/floci/gcp/test/GcsResumableUploadRawTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/floci/gcp/test/GcsResumableUploadRawTest.java
@@ -1,0 +1,159 @@
+package io.floci.gcp.test;
+
+import com.google.cloud.storage.BucketInfo;
+import com.google.cloud.storage.Storage;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class GcsResumableUploadRawTest {
+
+    private static final String BUCKET = TestFixtures.uniqueName("resumable-bucket");
+    private static final String OBJECT = "resumable.bin";
+    private static final int FIRST_CHUNK_SIZE = 16 * 1024 * 1024;
+    private static final int SECOND_CHUNK_SIZE = 1024 * 1024;
+    private static final int TOTAL_SIZE = FIRST_CHUNK_SIZE + SECOND_CHUNK_SIZE;
+    private static final HttpClient CLIENT = HttpClient.newHttpClient();
+
+    private static Storage storage;
+
+    @BeforeAll
+    static void setUp() {
+        storage = TestFixtures.storageClient();
+        storage.create(BucketInfo.of(BUCKET));
+    }
+
+    @AfterAll
+    static void tearDown() throws Exception {
+        if (storage instanceof AutoCloseable closeable) {
+            closeable.close();
+        }
+    }
+
+    @Test
+    void resumableUploadKeepsSessionOpenUntilFinalChunk() throws Exception {
+        HttpResponse<String> init = CLIENT.send(
+                HttpRequest.newBuilder(URI.create(TestFixtures.endpoint()
+                                + "/upload/storage/v1/b/" + BUCKET
+                                + "/o?uploadType=resumable&name=" + OBJECT))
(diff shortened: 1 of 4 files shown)


hectorvent:
Thanks again, the chunked resumable handling here is solid and I confirmed the happy path matches what the Java SDK expects (308 plus Range for non final chunks, 200 with size on finalize). But the `sdk-test-node` check is failing and it is caused by this change, so I think it is blocking.

The Node SDK's `file.save()` uses a resumable upload and sends `Content-Range: bytes 0-*/*`, with a star for the end offset. The new `parseContentRange` calls `parseLong` on that star and throws, so the upload returns 400 and the whole Node GCS suite cascades into 9 failures. The reason it is a regression: on main `resumablePut` ignores the Content-Range header and finalizes in one shot, so this request used to work. Java, Go, and Python pass because they send a numeric finalizing range, Node is the one using the star end form.

To fix it the parser needs to accept a star end offset (`bytes 0-*/*`), and just as importantly the default `save()` path has to still finalize so the object actually gets stored. While you are in there it would be worth also handling the size only finalize (`bytes */<size>`) and the status query (`bytes */*`) forms the SDKs use on resume. Capturing the exact request sequence the Node client sends, or running the `sdk-test-node` suite locally, should make the expected behavior clear. A test mirroring `file.save()` would guard against this going forward.

Everything else in the PR looks good, this is the one thing to sort out before merge.

hectorvent (approved):

PR #49, opened 2026-06-13
feat(cloudrun): add Docker-backed service execution

shirkevich:
## Summary

Adds experimental Docker-backed execution for Cloud Run v2 services while keeping the control-plane API compatible with the existing floci-gcp REST/LRO behavior.

The PR covers:

- Cloud Run v2 service create/get/list/patch/delete, revisions, IAM policy endpoints, and long-running operation polling.
- Optional execution mode that starts one Docker container for the latest ready revision.
- `FLOCI_GCP_SERVICES_CLOUDRUN_EXECUTION_MOCK=true` for metadata-only behavior when the execution flag is enabled but Docker runtime startup should be skipped.
- Deterministic Cloud Run-style app-root URLs on the floci-gcp front door.
- Backward-compatible invocation through `/run/v2/projects/{project}/locations/{location}/services/{service}`.
- Bounded create/update/delete operation handling so Terraform/OpenTofu replacement destroys do not hang on Docker cleanup.
- Bounded Docker API calls in the shared container lifecycle path, including Docker client reset on timeout/interruption.
- Force-removal cleanup for Cloud Run runtime containers, with runtime store cleanup in `finally` so stale runtime records are not left behind when Docker cleanup fails.
- Cloud Run GCS volume mounts backed by emulator GCS objects and Docker named volumes.
- SDK compatibility CI stabilization by running SDK jobs with Kafka mock mode, since those jobs assert Kafka management-plane behavior and do not need a live Redpanda broker.
- Compatibility workflow hardening so long-running test steps time out before the job and dump emulator logs on failures or cancellations.

## Behavior

Execution is disabled by default. When `FLOCI_GCP_SERVICES_CLOUDRUN_EXECUTION_ENABLED=true` and `FLOCI_GCP_SERVICES_CLOUDRUN_EXECUTION_MOCK=false`, create and template-changing patch operations start the requested workload image in Docker and update the service to Ready only after the runtime is reachable.

When `FLOCI_GCP_SERVICES_CLOUDRUN_EXECUTION_MOCK=true`, the Cloud Run Docker path is gated off with `!mock()` and services stay metadata-only even if execution is enabled. This mirrors the existing Kafka mock configuration shape.

While a create operation is still pending, the service metadata is persisted with `Ready=CONDITION_PENDING` and no `latestReadyRevision`. Clients should wait for the LRO before invoking the service.

Template-changing patch creates a new revision. The old ready revision remains invokable while the replacement container starts. Once the new runtime is ready, `latestReadyRevision` moves to the new revision and older runtime containers for that service are stopped.

PATCH accepts a Cloud Run v2 `Service` body and optional `updateMask`. `allow_missing` upsert behavior and etag preconditions are not implemented.

Delete removes service/revision metadata and completes the delete LRO before best-effort runtime cleanup, with cleanup bounded by the configured timeout. Runtime container cleanup uses Docker force remove semantics.

Runtime support is intentionally narrow:

- Exactly one container.
- Image is required.
- At most one container port.
- Plain env values are supported; env `valueSource` is rejected.
- Unsupported Cloud Run features fail before Docker startup instead of creating partial runtimes.

## Invocation URLs

Execution-mode service metadata returns deterministic host-routed URLs:

```text
http://{service}-{project-token}.{location}.run.{suffix}:{port}
```

For local defaults this is shaped like:

```text
http://orders-api-8af9d8563aae.us-central1.run.localhost.floci.io:4588
```

The generated host routes back through floci-gcp. A pre-matching filter resolves hosts in the `*.run.{suffix}` form, rewrites them internally to the existing invocation controller, and preserves the app path/query. For example:

```text
GET http://orders-api-....us-central1.run.localhost.floci.io:4588/api/database?x=1
```

reaches the runtime container as:

```text
/api/database?x=1
```

The proxy preserves useful external request metadata:

- `X-Forwarded-Host` is the generated Cloud Run host.
- `X-Forwarded-Uri` is the external app path/query, not the internal rewritten admin path.

The legacy invocation path remains supported for compatibility and focused tests:

```text
/run/v2/projects/{project}/locations/{location}/services/{service}
```

Direct random Docker container ports are internal implementation details and are not returned in Cloud Run service metadata.

## GCS Volumes

Cloud Run GCS volumes are supported for execution mode when declared as `template.volumes[].gcs` and mounted with container `volumeMounts`.

Implemented semantics:

- The referenced bucket must already exist in the floci-gcp GCS emulator before runtime startup.
- At startup, floci-gcp snapshots the bucket's current live objects into a Docker named volume.
- Docker volume mounts use `NoCopy` so workload image contents do not overwrite the materialized bucket snapshot.
- `readOnly=true` becomes a Docker read-only volume mount.
- `readOnly=false` allows writes inside the workload container.
- Writable mounts are synced back to emulator GCS when the runtime container is stopped during service delete or revision replacement.
- GCS object bodies are persisted separately from object metadata in `gcs-object-data.json`, so writable volume sync survives emulator restarts in persistent/hybrid/WAL storage modes.
- Writable sync mirrors regular files in the mounted volume back to the bucket prefix and removes previously existing objects under that prefix that are no longer present in the volume. This is data-loss-capable by design for writable mounts.
- Writes are not live-synced during request handling; other GCS clients observe them only after runtime cleanup.
- `subPath` is supported as an object-prefix path inside the bucket snapshot.
- GCS `mountOptions` are rejected.

Non-GCS volume sources are still unsupported in execution mode, including Secret, Cloud SQL, emptyDir, NFS, and other volume types.

The GCS snapshot/copy-back path uses a short-lived `alpine:3.20` helper container and Docker archive copy APIs. Docker must be able to use that image when GCS volumes are mounted.

## Docker Reliability

Cloud Run execution operations are capped by `FLOCI_GCP_SERVICES_CLOUDRUN_EXECUTION_OPERATION_TIMEOUT`. Docker cleanup after delete/replacement is best-effort and capped by `FLOCI_GCP_SERVICES_CLOUDRUN_EXECUTION_CLEANUP_TIMEOUT`.

Shared Docker lifecycle API calls are also capped by `FLOCI_GCP_DOCKER_API_TIMEOUT` (default `30s`). This applies to the container lifecycle manager path used for container create/start/inspect/remove, Docker volume inspect/create/remove, GCS volume archive copy, helper-container cleanup, and Kafka's local container inspect path. When a bounded Docker API call times out or is interrupted, floci-gcp cancels the call, resets the cached Docker Java client, and creates a fresh client for later calls instead of continuing to reuse a potentially poisoned client.

Docker volume inspect/create/remove now logs before and after the Docker API call. This makes hangs around GCS volume materialization visible in emulator logs.

## Usage Assumptions

For bare-host development, the generated URL may not resolve through local DNS. In that case either:

- Send traffic to `localhost:4588` with the generated `Host` header, or
- Use the legacy `/run/v2/projects/.../services/...` invocation path.

For Dockerized use, run floci-gcp with Docker socket access and put workload containers on a network that can route back to floci-gcp. The compatibility tests use this shape:

```text
FLOCI_GCP_SERVICES_CLOUDRUN_EXECUTION_ENABLED=true
FLOCI_GCP_BASE_URL=http://floci-gcp:4588
FLOCI_GCP_HOSTNAME=floci-gcp
FLOCI_GCP_SERVICES_DOCKER_NETWORK=<shared docker network>
```

Inside that Docker network, the embedded DNS path resolves generated `*.run.floci-gcp` hosts back to the floci-gcp container.

Clients should treat execution as an emulator feature, not a production-compatible Cloud Run runtime. It is intended to make local IaC and app smoke tests practical.

## Tested

Local Maven/JUnit:

```text
JAVA_HOME=/opt/homebrew/opt/openjdk@25/libexec/openjdk.jdk/Contents/Home ./mvnw test -Dtest=CloudRunExecutionRestIntegrationTest,ContainerLifecycleManagerTest,CloudRunRuntimeServiceTest
```

Result: 20 tests, 0 failures.

```text
JAVA_HOME=/opt/homebrew/opt/openjdk@25/libexec/openjdk.jdk/Contents/Home ./mvnw test -Dtest=GcsServiceTest
```

Result: 12 tests, 0 failures. This includes object-body persistence across a persistent store reload and cleanup of stale object metadata that has no object data.

```text
JAVA_HOME=/opt/homebrew/opt/openjdk@25/libexec/openjdk.jdk/Contents/Home ./mvnw test -Dtest=CloudRunExecutionRestIntegrationTest,CloudRunRuntimeServiceTest,ContainerLifecycleManagerTest,GcsServiceTest
```

Result: 32 tests, 0 failures.

```text
JAVA_HOME=/opt/homebrew/opt/openjdk@25/libexec/openjdk.jdk/Contents/Home ./mvnw test
```

Result: 244 tests, 0 failures.

Package and Docker verification:

```text
JAVA_HOME=/opt/homebrew/opt/openjdk@25/libexec/openjdk.jdk/Contents/Home ./mvnw clean package -DskipTests
docker build -f docker/Dockerfile.jvm-package -t floci-gcp:codex-cloud-run-execution .
```

Both passed.

Latest local image rebuild for the GCS persistence fix also passed:

```text
JAVA_HOME=/opt/homebrew/opt/openjdk@25/libexec/openjdk.jdk/Contents/Home ./mvnw clean package -DskipTests
docker build -f docker/Dockerfile.jvm-package -t floci-gcp:platform-integration -t floci-gcp:gcs-persistence-eb9b518 .
```

Resulting local image digest: `sha256:8df5776eaa39cb584414c23ec75860d67be0f90724d851a40548b3992293c57a`.

Dockerized GCS persistence restart smoke:

```text
IMAGE=floci-gcp:platform-integration /private/tmp/floci-gcs-persistence-smoke.sh
```

Result: passed. The smoke uploaded `mounted/smoke.txt`, restarted the emulator container with the same persistent data directory, verified `gcs-object-data.json` existed, and downloaded the original object body after restart.

Docker cleanup check after the final full-suite run showed only older pre-existing floci-labeled artifacts from previous runs; the current test-created Cloud Run runtime/helper containers and GCS volumes were cleaned up.

Downstream Atmos/OpenTofu integration slice:

```text
FLOCI_IMAGE=floci-gcp:codex-cloud-run-execution scripts/integration/floci-cloud-run-slice.sh
```

Result: passed.

This was run from an isolated temp copy of `atmos-gcp-platform` using the repo's Cloud Run component and the current `hashicorp/google` provider selection (`v7.36.0`). The clean full downstream script still reaches Cloud SQL before Cloud Run and fails with `405` because this branch does not include the Cloud SQL implementation; this slice removed Cloud SQL/PubSub state references only to isolate the Cloud Run path. It applied the GCS bucket and Cloud Run service, invoked the generated host URL `http://orders-api-8af9d8563aae.us-central1.run.localhost:4588`, force-replaced the same Cloud Run service name twice, wrote through the mounted GCS volume, and verified the mounted file was present after replacement. The replacement timings in that run were 10s destroy / 10s create for both forced replacements.

Focused Docker-backed Cloud Run integration coverage includes:

- Generated host-routed service URI invocation.
- Legacy `/run/v2/...` invocation path.
- Missing service not-found behavior.
- Read-only GCS volume mounted into nginx and served from app root.
- Writable GCS volume sync back to emulator GCS on service delete.
- Same-name Cloud Run service replacement with a GCS volume mount.
- Runtime cleanup using Docker force remove.
- Runtime store cleanup even when Docker cleanup throws.
- Mock-mode execution gate that avoids runtime startup.
- Retry-tolerant HTTP invocation assertions after LRO completion, because runtime readiness is TCP-based and app HTTP readiness can lag briefly.

Terraform/OpenTofu fixture checks from this PR cover provider-facing Cloud Run create/update/replacement and generated URI invocation paths.

## Known Limitations

- Execution mode remains experimental and opt-in.
- Single-container services only.
- No sidecars.
- No traffic splitting.
- No autoscaling or scale-to-zero.
- No Cloud Run auth/IAM invocation enforcement.
- No jobs or worker pools.
- No startup probes.
- No secrets/env value sources.
- No Cloud SQL volumes yet.
- No non-GCS volumes.
- No `allow_missing` or etag preconditions on update.
- No live GCS volume sync while the workload is running.
- Writable GCS volume sync mirrors the volume back to the bucket prefix, including deleting bucket objects that are absent from the mounted filesystem.
- Writable GCS volume sync currently stores copied-back files with `application/octet-stream`.
- GCS volume snapshot/copy-back reads object/file contents through emulator memory and Docker archive APIs, so this is intended for local smoke-test scale rather than large production buckets.


hectorvent:
Really like this — clean and well-tested. A few things to change before merge:

- **`runtime` config → `mock` boolean.** `runtime()` only accepts `"docker"` today (everything else throws), so it's effectively one value. We already do this for Kafka with `mock` (`FLOCI_GCP_SERVICES_KAFKA_MOCK`) — let's match it: `@WithDefault("false") boolean mock()` and gate the Docker path on `!mock()`.
- **Don't break `./mvnw test` without Docker.** `CloudRunExecutionRestIntegrationTest` needs a Docker daemon + image pulls with no guard, so the plain test run fails without Docker. The `mock` mode above, or an `assumeTrue(dockerAvailable)`, fixes it.
- **Two doc notes:** `update` doesn't do `allow_missing`/etag, and a writable GCS volume deletes bucket objects not in the volume on cleanup (data-loss-capable) — worth calling out.

Also, the SDK test jobs (go/java/node/python) are red on the latest run — mind taking a look? compat + build are green, so it looks isolated to those.

Rest looks great 👍

hectorvent:
@shirkevich please take a look at the conflicts.

hectorvent (approved):

hectorvent:
Thanks @shirkevich,

I appreciate your contribution.

PR #76, opened 2026-06-26
ci: run compatibility tests against native image

electrum:
Builds the native emulator image in compatibility CI and runs the existing compatibility matrix against it on native-architecture test containers. Registers Eventarc protobuf messages for native JSON handling.


hectorvent (approved):
Thank you @electrum

PR #82, opened 2026-07-02
fix(gcs): use snake_case field names in notification request and response

nicoechegut:
## Summary

The Google Storage REST API uses snake_case for notification fields (`payload_format`, `event_types`, `object_name_prefix`, `custom_attributes`), but `createNotification` was reading camelCase keys from the request body, silently dropping those values. The GET response also serialized as camelCase, so clients reading back a notification would receive empty values for those fields.

Fix `GcsService` to accept snake_case keys and add `@JsonProperty` annotations to `StoredNotification` so the response matches the API spec: https://docs.cloud.google.com/storage/docs/json_api/v1/notifications

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

Verified by creating a GCS Notification via a Terraform test and the `hashicorp/google` Terraform provider.
Before the fix, Terraform fails to apply with error, indicating the API returns empty fields:

```
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to
│ module.gcs_bucket.google_storage_notification.this[0], provider
│ "provider[\"registry.terraform.io/hashicorp/google\"]" produced an
│ unexpected new value: .payload_format: was cty.StringVal("JSON_API_V1"),
│ but now cty.StringVal("").
│ 
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.
```

Ran the sdk-test-java compatibility tests which include `GcsNotificationTest`.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/io/floci/gcp/test/GcsNotificationTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/floci/gcp/test/GcsNotificationTest.java
@@ -37,6 +37,7 @@ import org.junit.jupiter.api.TestMethodOrder;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
 import static org.assertj.core.api.Assertions.assertThat;
@@ -57,7 +58,7 @@ class GcsNotificationTest {
     private static TopicAdminClient topicAdminClient;
     private static SubscriptionAdminClient subscriptionAdminClient;
 
-    private static String notificationId;
+    private static Notification createdNotification;
 
     @BeforeAll
     static void setUp() throws IOException {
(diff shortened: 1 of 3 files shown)


hectorvent:
Thanks for the fix! 

Verified against the JSON API spec. `notifications` really is the one GCS resource using snake_case, so this is exactly right. It also fixes event-type/prefix filtering, which was silently ignored before.

Two small asks:

**1.** Add round-trip assertions to `GcsNotificationTest` (it only checks `topic`/`id`, so it can't catch a regression here):

```java
assertThat(notification.getPayloadFormat()).isEqualTo(NotificationInfo.PayloadFormat.JSON_API_V1);
assertThat(notification.getEventTypes())
        .containsExactlyInAnyOrder(NotificationInfo.EventType.OBJECT_FINALIZE,
                NotificationInfo.EventType.OBJECT_DELETE);
```

**2.** Add `@JsonAlias` so existing persisted data (camelCase) still loads:

```java
@JsonProperty("payload_format")
@JsonAlias("payloadFormat")
private String payloadFormat = "JSON_API_V1";
```

Otherwise LGTM

nicoechegut:
Thanks for the review @hectorvent
I've addressed the feedback: added round-trip assertions and Json alias on all notification fields.

hectorvent (approved):
Thanks @nicoechegut

PR #91, opened 2026-07-15
feat(iamcredentials): add generateAccessToken support

snazy:
Adds the IAM Credentials REST emulator surface for local credential vending, including generateAccessToken validation, service enablement wiring, REST tests, and Java generated-client compatibility coverage.

- [x] New feature (`feat:`)

Implements the IAM Credentials v1 HTTP/JSON generateAccessToken path used by the Google Java IAM Credentials client. Verified with IamCredentialsServiceTest, IamCredentialsRestIntegrationTest, IamCredentialsDisabledRestIntegrationTest, IamServiceTest, and compatibility-tests/sdk-test-java IamCredentialsTest against a running emulator.

- [x] `./mvnw test` passes locally
- [x] New or updated integration test added
- [x] Commit messages follow Conventional Commits

PR #92, opened 2026-07-25
feat(gcs): support service account authentication

electrum:
Allows GCS clients to use standard service-account credentials against the emulator.

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/io/floci/gcp/test/GcsServiceAccountAuthenticationTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/floci/gcp/test/GcsServiceAccountAuthenticationTest.java
@@ -1,0 +1,48 @@
+package io.floci.gcp.test;
+
+import com.google.auth.oauth2.ServiceAccountCredentials;
+import com.google.cloud.storage.BucketInfo;
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageOptions;
+import org.junit.jupiter.api.Test;
+
+import java.net.URI;
+import java.security.KeyPair;
+import java.security.KeyPairGenerator;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class GcsServiceAccountAuthenticationTest {
+
+    @Test
+    void serviceAccountCredentialsAuthenticateStorageRequests() throws Exception {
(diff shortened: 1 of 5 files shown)


hectorvent:
Thanks David, this is a nice, tightly scoped addition. Pointing the credential's `token_uri` at the emulator so the SDK runs its real OAuth flow is exactly the right shape, and the wire contract matches what `ServiceAccountCredentials.refreshAccessToken()` sends and parses ([google-auth-library-java](https://github.com/googleapis/google-auth-library-java/blob/main/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java)).

1. **(blocking)** `/token` stands in for `https://oauth2.googleapis.com/token`, the platform-wide token server every credential uses ([OAuth2Utils](https://github.com/googleapis/google-auth-library-java/blob/main/oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java) hardcodes it as the default), but it is registered through `GcsService`, so it disappears when GCS is disabled. Would you consider hoisting it to a shared location, and aligning with the token-vending work in #90 and #91 so we end up with one token-minting story?

2. **(follow-up, separate PR)** Per [RFC 6749 §5.2](https://datatracker.ietf.org/doc/html/rfc6749#section-5.2), an unrecognized grant type should return `unsupported_grant_type`, with `invalid_grant` reserved for a bad assertion.

3. **(follow-up, separate PR)** The compat test pins the internal `floci-gcp-` token prefix; asserting a non-blank token that later Storage calls accept would be less brittle.

Docs read well, thank you!

electrum:
@hectorvent Addressed the blocking point and rebased. The `/token` endpoint is now an application-scoped `OAuthTokenController` under the shared `services.credentials` package, rather than being registered through `GcsService`, so it is available independently of whether GCS is enabled.

I also added `OAuthTokenGcsDisabledRestIntegrationTest`, which disables GCS and verifies that the JWT bearer exchange still succeeds.

electrum:
@hectorvent I also folded both follow-up items into this PR. Unsupported grant types now return `unsupported_grant_type`, while a missing or blank JWT assertion returns `invalid_grant`, with separate REST coverage for each case. The REST and Java SDK tests now require only a non-blank access token instead of depending on the internal `floci-gcp-` token prefix.

hectorvent (approved):
Thanks @electrum

PR #93, opened 2026-07-25
fix(gcs): preserve URI-encoded object names

electrum:
Cloud Storage object names are opaque strings. URI-looking sequences such as
`%2F`, `%20`, `%25`, and `%2B` remain part of the object name; they are not
decoded a second time into `/`, a space, `%`, or `+`.

For an object named `literal%2Fslash`, the percent sign must be URI-encoded as
`%25`. The HTTP request target therefore contains `literal%252Fslash`. GCS
decodes the URI once and returns the original object name, including the
literal `%2F`.

Real GCS preserves the object name when using `gcloud`:

```console
$ printf 'encoded object\n' > gcs-log-http-example.txt
$ gcloud storage cp gcs-log-http-example.txt \
    gs://floci-gcloud-log-http-20260725-085458-47f02e/literal%2Fslash
Copying file://gcs-log-http-example.txt to gs://floci-gcloud-log-http-20260725-085458-47f02e/literal%2Fslash

$ gcloud storage ls 'gs://floci-gcloud-log-http-20260725-085458-47f02e/**'
gs://floci-gcloud-log-http-20260725-085458-47f02e/literal%2Fslash

$ gcloud storage cat \
    gs://floci-gcloud-log-http-20260725-085458-47f02e/literal%2Fslash
encoded object
```

The `gcloud --log-http` output exposes the actual JSON API request and
response. The authorization token is redacted by `gcloud`; unrelated headers
and response fields are omitted below:

```console
$ gcloud storage objects describe \
    gs://floci-gcloud-log-http-20260725-085458-47f02e/literal%2Fslash \
    --format='json(bucket,name,size)' \
    --log-http
=======================
==== request start ====
uri: https://storage.googleapis.com/storage/v1/b/floci-gcloud-log-http-20260725-085458-47f02e/o/literal%252Fslash?alt=json&projection=full
method: GET
== headers start ==
b'accept': b'application/json'
b'authorization': --- Token Redacted ---
== headers end ==
==== request end ====
---- response start ----
status: 200
-- headers start --
Content-Type: application/json; charset=UTF-8
-- headers end --
-- body start --
{
  "kind": "storage#object",
  "id": "floci-gcloud-log-http-20260725-085458-47f02e/literal%2Fslash/1784994917658262",
  "selfLink": "https://www.googleapis.com/storage/v1/b/floci-gcloud-log-http-20260725-085458-47f02e/o/literal%252Fslash",
  "mediaLink": "https://storage.googleapis.com/download/storage/v1/b/floci-gcloud-log-http-20260725-085458-47f02e/o/literal%252Fslash?generation=1784994917658262&alt=media",
  "name": "literal%2Fslash",
  "bucket": "floci-gcloud-log-http-20260725-085458-47f02e",
  "size": "15"
}
-- body end --
---- response end ----
----------------------
{
  "bucket": "floci-gcloud-log-http-20260725-085458-47f02e",
  "name": "literal%2Fslash",
  "size": 15
}
```

This preserves the same behavior across JSON, XML, download, and batch
operations, so clients address the object name they originally supplied.

### DIFF
--- a/compatibility-tests/sdk-test-gcloud/test/storage.bats
+++ b/compatibility-tests/sdk-test-gcloud/test/storage.bats
@@ -44,6 +44,20 @@ setup() {
     assert_output --partial "gs://${BUCKET}/data.txt"
 }
 
+@test "storage: preserve URI-encoded object names" {
+    local f="${BATS_TEST_TMPDIR}/encoded.txt"
+    echo "encoded object" > "$f"
+
+    for object in "literal%2Fslash" "literal%20space" "literal%25percent" "literal%2Bplus"; do
+        run gcloud_cmd storage cp "$f" "gs://${BUCKET}/${object}"
+        assert_success
+
+        run gcloud_cmd storage cat "gs://${BUCKET}/${object}"
+        assert_success
+        assert_output --partial "encoded object"
+    done
+}
+
 @test "storage: delete object" {
     local f="${BATS_TEST_TMPDIR}/del.txt"
     echo "x" > "$f"
--- a/compatibility-tests/sdk-test-java/src/test/java/io/floci/gcp/test/GcsBatchTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/floci/gcp/test/GcsBatchTest.java
@@ -27,6 +27,7 @@ class GcsBatchTest {
     private static final String OBJ1 = "batch/object1.txt";
     private static final String OBJ2 = "batch/object2.txt";
     private static final String OBJ3 = "batch/object3.txt";
+    private static final String ENCODED_OBJECT = "batch/literal%2Fname.txt";
 
     private static Storage storage;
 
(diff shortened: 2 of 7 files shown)


hectorvent:
Thank you for this. The `--log-http` capture against real GCS in the description is exactly the kind of evidence that makes a wire-format fix easy to review, and the three-layer test coverage (raw REST with URL encoding disabled, gcloud bats, Java SDK batch) is much appreciated.

The diagnosis looks right to me: JAX-RS already decodes path params once, so the extra `URLDecoder.decode` was a second decode. And since `URLDecoder` is form-decoding, it was also silently turning `+` into a space, which your `literal%2Bplus` case covers nicely. Single decode at the HTTP layer matches what the JSON API documents for `o/{object}` ([discovery doc](https://www.googleapis.com/discovery/v1/apis/storage/v1/rest), [encoding URI path parts](https://cloud.google.com/storage/docs/request-endpoints#encoding)).

The batch change is a nice catch too. The multi-arg `URI` constructor treats its components as unencoded, so passing `getRawPath()` through it was re-encoding `%` into `%25` and double-encoding subrequest targets. Raw-string concatenation is the right call there.

One small observation, entirely optional for this PR: `GcsBucketController` still routes its ACL `entity` path params through the same `URLDecoder.decode` pattern, so an entity like `user-name+tag@example.com` would hit the same plus-to-space corruption on the bucket/default-ACL endpoints. Since this PR fixed the object-ACL side, it might be worth a quick follow-up to bring the bucket side in line. Happy to see that as a separate PR.

Looks good, thank you!

hectorvent (approved):

PR #94, opened 2026-07-25
fix(gcs): enforce object preconditions atomically

electrum:
Prevents concurrent GCS writes from both succeeding when they use the same object precondition.

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/io/floci/gcp/test/GcsComposeTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/floci/gcp/test/GcsComposeTest.java
@@ -5,6 +5,7 @@ import com.google.cloud.storage.BlobId;
 import com.google.cloud.storage.BlobInfo;
 import com.google.cloud.storage.BucketInfo;
 import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageException;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
@@ -12,6 +13,7 @@ import org.junit.jupiter.api.Test;
 import java.nio.charset.StandardCharsets;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 /**
  * Regression coverage for issue #1: object compose was broken
@@ -62,5 +64,21 @@ class GcsComposeTest {
                 StandardCharsets.UTF_8);
         assertThat(content).isEqualTo("hello world");
         assertThat(composed.getSize()).isEqualTo("hello world".getBytes(StandardCharsets.UTF_8).length);
+    }
+
+    @Test
+    void composeDoesNotExistPreconditionFailsWith412() {
+        storage.create(BlobInfo.newBuilder(BlobId.of(BUCKET_NAME, "precondition-source")).build(), new byte[0]);
+        BlobInfo target = BlobInfo.newBuilder(BlobId.of(BUCKET_NAME, "precondition-target")).build();
+        storage.create(target, new byte[0]);
+        Storage.ComposeRequest request = Storage.ComposeRequest.newBuilder()
+                .addSource("precondition-source")
+                .setTarget(target)
+                .setTargetOptions(Storage.BlobTargetOption.doesNotExist())
+                .build();
+
+        assertThatThrownBy(() -> storage.compose(request))
+                .isInstanceOf(StorageException.class)
+                .satisfies(e -> assertThat(((StorageException) e).getCode()).isEqualTo(412));
     }
 }
--- a/compatibility-tests/sdk-test-java/src/test/java/io/floci/gcp/test/GcsCopyRewriteTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/floci/gcp/test/GcsCopyRewriteTest.java
@@ -6,6 +6,7 @@ import com.google.cloud.storage.BlobInfo;
 import com.google.cloud.storage.BucketInfo;
 import com.google.cloud.storage.CopyWriter;
 import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageException;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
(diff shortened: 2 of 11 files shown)


electrum:
Regarding the process-local object locks: this matches Floci's supported storage model. `InMemoryStorage`, `PersistentStorage`, `HybridStorage`, and `WalStorage` each keep their authoritative state in a process-local `ConcurrentHashMap`; their synchronization and persistence coordination are process-local as well. Multiple emulator processes sharing the same backing files would already race for all stored resources, independently of GCS object preconditions, and `StorageBackend` does not expose a cross-process transaction or locking primitive.

In the supported one-process-per-storage deployment, `GcsService` is application-scoped and all operations for an object use the same striped lock, so the precondition check and mutation are atomic. Adding distributed or file locking only to GCS would imply a shared-storage guarantee that the storage layer does not provide.

The separate restart-generation concern was valid and is fixed in c6d5955 by initializing the sequence above the highest persisted generation.

hectorvent:
Thanks David, this is a really clean fix. Moving the precondition check inside a striped per-object lock and replacing the `currentTimeMillis()` generation with a seeded monotonic sequence kills two real races at once (check-then-act between concurrent writers, and same-millisecond writes colliding on generation), and re-checking the stored preconditions when a resumable upload finalizes is a nice touch. The concurrent SDK test asserting exactly one winner out of eight writers is exactly the kind of coverage this needed.

A few observations, none blocking:

1. `objects.delete` also accepts `ifGenerationMatch` / `ifGenerationNotMatch` / `ifMetagenerationMatch` / `ifMetagenerationNotMatch` per the [discovery doc](https://www.googleapis.com/discovery/v1/apis/storage/v1/rest), and the Java SDK exposes them via `BlobSourceOption.generationMatch()` on delete, but our DELETE endpoint currently ignores them (as do compose/copyTo/rewriteTo destinations). Not this PR's fault, and with `GcsObjectPreconditions` and the lock now in place it would slot in naturally. Totally fine as a follow-up, just flagging while we're in the area.

2. Adjacent pre-existing quirk in `checkPreconditions` (which this PR moves): for `ifGenerationNotMatch` the spec says "If no live object exists, the precondition fails", but we currently pass in that case (see [request preconditions](https://cloud.google.com/storage/docs/request-preconditions)). Happy to take that as a separate fix if you'd rather keep this PR scoped to atomicity.

3. Question on resumable uploads: you now enforce preconditions both at session initiation and again at finalize. Do you happen to know whether real GCS 412s at initiation, or only when the upload completes? The finalize re-check is definitely right; I'm just curious whether the early check could reject a session real GCS would have accepted (and let fail later). Benign either way for SDK behavior, so no action needed unless you know the answer offhand.

Looks good, thank you!

hectorvent (approved):

hectorvent:
@electrum Please resolve conflicts when you have a chance.

electrum:
@hectorvent Thanks, rebased

electrum:
Thanks for flagging these. All three are addressed in the updated commit.

1. Delete now accepts and atomically enforces all four object preconditions. Copy and rewrite atomically enforce all four destination preconditions. Compose enforces the two destination preconditions supported by the JSON API, `ifGenerationMatch` and `ifMetagenerationMatch`.

2. `ifGenerationNotMatch` now fails with 412 when no live object exists. The missing-object handling for metageneration preconditions was corrected at the same time.

3. I verified the resumable behavior against real GCS. It accepts the session even when the precondition is already false, then enforces the precondition when the upload is finalized:

```http
POST /upload/storage/v1/b/<bucket>/o?uploadType=resumable&name=already-exists&ifGenerationMatch=0
Authorization: Bearer <token>
Content-Type: application/json
X-Upload-Content-Type: text/plain

{}
```

```http
HTTP/2 200
location: https://storage.googleapis.com/upload/storage/v1/b/<bucket>/o?...&upload_id=<upload-id>
content-length: 0
```

Finalizing that session returns:

```http
PUT <location>
Content-Type: text/plain

resumable payload
```

```http
HTTP/2 412
content-type: application/json; charset=UTF-8

{
  "error": {
    "code": 412,
    "message": "At least one of the pre-conditions you specified did not hold.",
    "errors": [
      {
        "reason": "conditionNotMet"
      }
    ]
  }
}
```

I also repeated the test by initiating the session while the object was absent, creating a competing object, and then finalizing the session. Initiation returned 200 and finalization returned 412. Floci now matches this by retaining the preconditions with the session and evaluating them only during the atomic finalization.

hectorvent:
Thanks for the quick turnaround, and for going past what I asked. I verified all three against the source: `ifGenerationNotMatch` on a missing live object now fails, per the [discovery doc](https://www.googleapis.com/discovery/v1/apis/storage/v1/rest). The new SDK tests are a nice addition.

**(blocking)** Logistics only, nothing wrong with the patch. #120 just merged and collides with this: both add a component to `ResumableUpload` in the same position, so it went conflicting when that landed. Would you rebase? The union is `(bucket, objectName, contentType, customerEncryption, metadata, preconditions, data)`. Sorry for the ordering, their CI is gated on every push.

**(follow-up, separate PR)** One narrow case: when `generation` addresses an archived revision, preconditions are evaluated against that version, not the live object. The doc frames them around the live one ("if no live object exists, the precondition fails"), but I could not settle the archived case, so treat it as a question.

Happy to merge once it is green.

electrum:
Rebased onto current `main` after #120. The conflict was resolved by carrying both custom metadata and preconditions through resumable uploads as `(bucket, objectName, contentType, customerEncryption, metadata, preconditions, data)`, preserving both behaviors.

The repeated Greptile cross-process concern is resolved by Floci's supported storage model. All storage backends keep authoritative state in one process, and sharing backing files between emulator processes is unsupported and already unsafe for all resources. Within the supported one-process-per-storage deployment, the application-scoped GCS service uses one striped lock set, so the precondition check and mutation are atomic. The detailed rationale is in [this earlier comment](https://github.com/floci-io/floci-gcp/pull/94#issuecomment-5079170435).

The full test suite passes: 658 tests, 0 failures, 0 errors, 0 skipped.

hectorvent:
Thanks for the quick rebase. I verified the resolution: both custom metadata and preconditions flow through every resumable path, exactly the union above, and CI is fully green at the new head. The blocker above is cleared, no blockers from my side.

hectorvent (approved):

PR #96, opened 2026-07-27
feat(gke): model node pools as real resources, complete ClusterManage…

### LINKED ISSUE #95: [FEAT] GKE: model node pools as real resources + full ClusterManager RPC coverage
## Service

GKE (Kubernetes Engine)

## API Method / Feature

Node pools as real, independently-lifecycled resources (`CreateNodePool`, `GetNodePool`, `ListNodePools`, `DeleteNodePool`, `UpdateNodePool`), plus the rest of the `container.v1` `ClusterManager` mutation/read surface that's currently missing: `SetNodePoolAutoscaling`, `SetNodePoolManagement`, `SetNodePoolSize`, `UpdateCluster`, `SetResourceLabels`, `SetMasterAuth`, `SetNetworkPolicy`, `SetAddonsConfig`, `SetLoggingService`, `SetMonitoringService`, `SetLocations`, `SetLegacyAbac`, `SetMaintenancePolicy`, `StartIPRotation`/`CompleteIPRotation`, `CompleteNodePoolUpgrade`/`RollbackNodePoolUpgrade`, `GetServerConfig`, `GetJSONWebKeys`, `ListUsableSubnetworks`, `CheckAutopilotCompatibility`, `FetchClusterUpgradeInfo`/`FetchNodePoolUpgradeInfo`.

## GCP Documentation

https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.locations.clusters

## Why is this needed?

Right now node pools are a static, hardcoded field on the cluster response with no lifecycle of their own — there's no way to create, get, list, or delete a node pool independently of the cluster. That means the standard way almost everyone provisions GKE with Terraform/OpenTofu doesn't work against floci-gcp at all:

```hcl
resource "google_container_cluster" "this" {
  remove_default_node_pool = true
  initial_node_count       = 1
}

resource "google_container_node_pool" "default" {
  cluster = google_container_cluster.this.name
  autoscaling { min_node_count = 1; max_node_count = 3 }
  node_config { machine_type = "e2-medium" }
  management  { auto_repair = true; auto_upgrade = true }
}
```

I ran into this trying to point a real production GKE Terraform module at floci-gcp for local testing — `remove_default_node_pool` has nothing to delete against (404), and the standalone `google_container_node_pool` resource has nothing to create against either. Since docs/services/gke.md already flags this explicitly ("node pools are not modeled" / "the Terraform/OpenTofu google_container_cluster resource is not supported"), it seems like a known gap rather than something intentionally left out — just filing it so it's tracked.

- [x] Yes
- [ ] No

avison9:
### Summary

GKE cluster support existed, but node pools were a static hardcoded field, not
a real resource  so the standard `remove_default_node_pool = true` +
standalone `google_container_node_pool` Terraform pattern (the most common
real-world way to provision GKE) could not work at all. This PR adds full
node pool CRUD and the rest of the `container.v1` `ClusterManager` RPC
surface (mutation RPCs Terraform/gcloud rely on, IP rotation, node pool
upgrade acknowledgment, legacy ABAC, maintenance policy, server config, JWKS,
usable subnetworks, Autopilot compatibility checks, upgrade info)  the only
RPC intentionally left out is `CancelOperation`, since every operation here
is synchronous and there's never anything in flight to cancel. Fixes two
bugs along the way (found via live `terraform apply` testing) that would
otherwise cause perpetual cluster replacement on every `plan`.

Closes #95

```
# Verified end to end against a live instance:
$ terraform apply
Apply complete! Resources: 2 added, 0 changed, 0 destroyed.

$ terraform plan
No changes. Your infrastructure matches the configuration.
```

### Type of change

- [x] New feature (`feat:`)

### What changed

- `StoredNodePool` (new): node pools as real, independently-lifecycled resources.
- `GkeService`: node pool CRUD (`create/get/list/delete/updateNodePool`,
  `setNodePoolAutoscaling/Management/Size`) and cluster mutation RPCs
  (`updateCluster`, `setLabels`, `setMasterAuth`, `setNetworkPolicy`,
  `setAddonsConfig`, `setLoggingService`, `setMonitoringService`,
  `setLocations`). Cluster deletion cascades to its node pools.
- `KubernetesController`: matching REST routes, including GCP's colon-suffixed
  custom-method paths (`:setNetworkPolicy` etc.), using the existing
  `{param: [^:/]+}` JAX-RS pattern already established elsewhere in this repo.
- `StoredCluster`: typed fields for everything Terraform/gcloud actually diff
  on (`initialNodeCount`, `locations`, `labelFingerprint`, etc.), plus an
  `extraConfig` passthrough map for nested blocks the emulator doesn't act on
  semantically (`privateClusterConfig`, `workloadIdentityConfig`, etc.)  
  stored and echoed verbatim for exact round-trip fidelity.
- Also added: `setLegacyAbac`, `setMaintenancePolicy`, `startIpRotation`/
  `completeIpRotation`, `completeNodePoolUpgrade`/`rollbackNodePoolUpgrade`,
  `getServerConfig`, `getJsonWebKeys`, `listUsableSubnetworks`,
  `checkAutopilotCompatibility`, and `fetchClusterUpgradeInfo`/
  `fetchNodePoolUpgradeInfo`  the remaining standalone mutation/read RPCs.
  The five read-only ones return honest stub data (empty JWKS, a synthetic
  subnetwork, no compatibility issues, current-version-as-target) rather than
  fabricated analysis, since floci-gcp has no real infrastructure behind them.
- `KubernetesProjectController` (new): `ListUsableSubnetworks` is
  project-scoped with no `location` segment, unlike every other GKE method,
  so it needed its own controller mounted at `/container/v1/projects/{project}`.
- Confirmed Autopilot mode (`autopilot.enabled`) and Fleet/Anthos registration
  (`fleet`) already round-trip correctly through the existing `extraConfig`
  passthrough  no new code needed, just a regression test locking it in.

### Two bugs found and fixed via live Terraform validation

1. `initial_node_count` was never echoed back on the Cluster response, so the
   (ForceNew) attribute always read as `0` and the provider planned to
   destroy+recreate the cluster on every single `plan`, not just the first.
2. `network`/`subnetwork` are read by the provider from
   `Cluster.NetworkConfig.network`/`.subnetwork`, not the deprecated top-level
   fields  real GKE populates both locations; floci-gcp only set the
   top-level ones.

Also documented (not fixable here): the `hashicorp/google` provider's
`RemoveBasePathVersion` helper strips the last path segment of any custom
endpoint, so `container_custom_endpoint = "<endpoint>/container/"` silently
collapses to `<endpoint>/` and misroutes to Managed Kafka (near-identical REST
shape). GKE needs **host-mode** routing
(`container_custom_endpoint = "http://container.localhost:4588/"`), not path-mode
 now documented with a warning callout in `docs/services/gke.md`.

### GCP Compatibility

Traced the real HTTP method+path bindings and message shapes directly from
`googleapis/googleapis`'s `google/container/v1/cluster_service.proto` (the
authoritative source per this repo's own `AGENTS.md`), not just the Terraform
provider's behavior.

### How it was tested

- `./mvnw test`: 465 run, 0 failures, 0 errors (31 GkeServiceTest cases now,
  up from 6, all new behavior covered).
- **Live**: `./mvnw quarkus:dev` + real `terraform apply/plan/destroy` against
  a config mirroring a real-world production GKE module exactly
  (VPC-native networking, private cluster config, workload identity, master
  authorized networks, `remove_default_node_pool` + standalone node pool with
  autoscaling/node_config/management)  apply succeeds, a second `plan` shows
  zero drift, destroy is clean, and `remove_default_node_pool` was confirmed
  via direct `curl` to have actually deleted `default-pool`.
- Not yet wired into the CI `compatibility-tests/compat-terraform` suite 
  that requires a Docker network alias for `container.floci-gcp` in
  `compatibility.yml`, flagged as a follow-up rather than touched here to
  avoid risking the shared CI networking other suites depend on.

### Checklist


hectorvent:
Thank you for this, modeling node pools as real resources is exactly what the standalone `google_container_node_pool` pattern needed, and the proto grounded writeup with live Terraform evidence made it a pleasure to verify. I checked every new route against `cluster_service.proto` and the verbs, paths and field names all line up.

1. **(blocking)** `CompleteNodePoolUpgrade` returns `google.protobuf.Empty`, not an Operation ([cluster_service.proto#L382-L388](https://github.com/googleapis/googleapis/blob/aa87617d67/google/container/v1/cluster_service.proto#L382-L388)). Would you be open to returning `{}` from `:completeUpgrade`? (`rollback` really does return an Operation, which you have right.)

2. **(blocking)** `getCluster`/`listClusters` call `setNodePools` on the live stored object, so a stale pool snapshot gets persisted and the startup migration can resurrect deleted pools after a restart in persistent mode. Attaching pools on a copy (or at the JSON building layer) would close that hole.

3. **(follow-up, separate PR)** `:updateMaster` ([L221-L231](https://github.com/googleapis/googleapis/blob/aa87617d67/google/container/v1/cluster_service.proto#L221-L231)) looks like the one other missing RPC. Happy to track it separately, or the "full surface" wording could just mention it.

Really solid contribution, thank you!

avison9:
Thank you for the careful read. Checking every route against `cluster_service.proto` is exactly the review this needed, and both blocking points were real. Both are fixed, along with two findings Greptile raised on the latest push.

**1. `CompleteNodePoolUpgrade` returns `Empty`.** You're right, and I had it wrong. It now returns `{}` and no longer creates an Operation; `requireNodePool` still runs, so an unknown pool is still a `404`. `:rollback` is untouched, since it genuinely does return an Operation.

One deliberate leftover: `OperationType.COMPLETE_NODE_POOL_UPGRADE` is now unused, but I left the constant in place rather than deleting it. Operations are persisted through `StorageFactory`, so removing the enum value would break Jackson deserialization of any operation record written by an earlier build in `persistent` mode. Happy to remove it if you'd rather take that.

**2. `setNodePools` on the live stored object.** Good catch, and the consequence was worse than a stale read: the mutated record is what gets flushed, so in `persistent` mode a deleted pool could come back on the next startup via the embedded copy. `getCluster` and `listClusters` now return a detached copy (new `StoredCluster` copy constructor), and the read path no longer writes. Covered by `readingAClusterDoesNotWriteNodePoolsBackOntoTheStoredRecord`, which asserts against the store rather than the returned object, so it fails if anything reintroduces the mutation.

**3. `:updateMaster`.** Agreed, and I've left it out of this PR as you suggested. I'd rather not stretch the wording to cover it, so I can either open a follow-up issue for it now or send a small separate PR, whichever you prefer. Say the word.

Two further fixes from Greptile's latest pass, both worth flagging since they touch persistence:

* **Interrupted migration was non-resumable.** The migration skipped a cluster entirely once its pool store was non-empty, so a run that died partway never finished, and the remaining embedded pools were dropped for good. Pools now migrate individually, and the embedded list is cleared once a cluster is done. That clear is deliberately the "migrated" marker instead of "the store has pools for this cluster", because the latter is erased by a later `DeleteNodePool`, which is the same resurrection hole as your point 2 reached from the other side.
* **`desiredNodeVersion` only updated the cluster aggregate**, leaving each pool reporting its old `version`, so a reconciliation client would see the cluster upgraded and its pools not. It now updates the pools too.

Greptile's three earlier P1s (partial cluster creation, `@JsonIgnore` dropping pools, typed `desired*` fields overwritten by the `extraConfig` merge) were already addressed in `fix(gke): address review findings on node pool creation and updates`. They are stale comments from the first review pass rather than open items.

The branch has also been rebased onto current `main`, since it had drifted 42 commits behind and was showing conflicts. Only `README.md` needed hand resolution, where I kept your newer Cloud Run and Eventarc rows and took only the GKE row from this PR.

Full suite green: 882 tests, 0 failures. Three regression tests added for the above.

avison9:
On your third point: opened #177 to track `:updateMaster` separately, rather than widening this PR or stretching the "full surface" wording to cover an RPC that is not there.

I offered to pick it up, but after #96 lands, since it would otherwise conflict with the node pool work on this branch. The issue notes one detail worth pinning with a test when it is implemented: a master upgrade should leave `currentNodeVersion` and the node pools untouched, since real GKE upgrades the control plane independently of node pools.

hectorvent:
Thanks for the thorough round and for opening #177. Re-checked at `ff3196a`: both earlier blocking points are closed, with good tests for each.

1. **(blocking)** One thing the version propagation introduced: `desired_node_version` targets the pool named by `desired_node_pool_id`, which is "mandatory if desired_node_version ... is specified and there is more than one node pool" ([cluster_service.proto#L3561-L3565](https://github.com/googleapis/googleapis/blob/aa87617d67/google/container/v1/cluster_service.proto#L3561-L3565)). Terraform sends `DesiredNodePoolId: "default-pool"` ([resource_container_cluster.go#L4629-L4636](https://github.com/hashicorp/terraform-provider-google/blob/4de3c23/google/services/container/resource_container_cluster.go#L4629-L4636)), so upgrading every pool reads as drift on the other `google_container_node_pool` resources. Would you key on `desiredNodePoolId`, fall back to the sole pool, and 400 when it is absent with several pools? Greptile's framing pointed the other way, so an easy one to land on.

2. **(follow-up, separate PR)** `current_node_version` is "the minimum version of all nodes" ([#L3086-L3091](https://github.com/googleapis/googleapis/blob/aa87617d67/google/container/v1/cluster_service.proto#L3086-L3091)), which would also answer Greptile's UpdateNodePool point. And `docs/services/gke.md:187` still says completeUpgrade returns the operation.

Point 1 is the only thing left before merge.

avison9:
Fixed in `7b74bbd`. You were right, and the proto is unambiguous once you put the two fields side
by side — `desired_node_pool_id` is "mandatory if `desired_node_version` ... is specified and there
is more than one node pool", so applying the version to every pool was upgrading pools the request
never named.

`updateCluster` now resolves a target before it assigns anything:

- `desiredNodePoolId` given, that pool, `404` if it does not exist.
- Omitted with exactly one pool, that pool. This is the case the proto carves out, and it is what
  keeps the existing single-pool behaviour working.
- Omitted with several pools, `400`. There is no defensible target, so it is a bad request rather
  than a guess.

Resolution deliberately happens **before** `setCurrentNodeVersion`. Both rejection paths throw, and
`cluster` is the live stored object, so validating afterwards would leave a rejected request's
version behind on it — the same shape as the partial-creation finding from the earlier round.
`desiredNodeVersionNeedsAPoolIdOnceThereIsMoreThanOnePool` asserts the cluster and every pool are
untouched after the rejection, not just that it threw.

**One thing I found while wiring this up that you did not ask about.** `desiredNodePoolId` was
falling through the generic `desired*` merge into `extraConfig`, so it was serialized back as a
`nodePoolId` field the Cluster resource does not have. Terraform sends the field on every cluster
update, so this was firing constantly rather than being theoretical. It is one line
(`stripped.remove("nodePoolId")`) next to the existing removals for the other consumed typed
fields, and `desiredNodePoolIdDoesNotLeakIntoClusterState` covers it. Happy to split it out if you
would rather keep this commit to the one finding.

I also corrected two claims in `docs/services/gke.md` that this branch had already invalidated: the
"all mutations return an Operation" line and the `CompleteNodePoolUpgrade` sentence both still said
it returns the operation, which stopped being true when I changed it to `Empty` in `4b46d51`. That
is the doc line from your point 2. The `current_node_version` minimum-version semantics is the
other half of that point and is genuinely a separate change, so I have left it for the follow-up
along with Greptile's `UpdateNodePool` point, which it subsumes.

|---|---|---|---|---|---|
| Baseline, `ff3196a` | 883 | 0 | 0 | 0 | BUILD SUCCESS |
| With this change | 887 | 0 | 0 | 0 | BUILD SUCCESS |

Five tests added, one existing test renamed: `desiredNodeVersionUpdatesTheNodePoolsAsWellAsTheCluster`
is now `desiredNodeVersionUpdatesTheSolePoolWhenNoPoolIdIsGiven`, since it was asserting the
behaviour this change narrows and its old name would have been actively misleading.

Confirmed the tests detect the pre-fix behaviour rather than merely passing. Reverting the target
resolution to the old "every pool" fails three of them, including the drift you described:

```
[ERROR] desiredNodeVersionUpgradesOnlyThePoolNamedByDesiredNodePoolId:599
        a pool the request did not name must keep its version
        ==> expected: <1.30.0-gke.1> but was: <1.31.5-gke.1>
[ERROR] desiredNodeVersionNeedsAPoolIdOnceThereIsMoreThanOnePool:612
        Expected GcpException to be thrown, but nothing was thrown.
[ERROR] desiredNodePoolIdMustNameAnExistingPool:632
        Expected GcpException to be thrown, but nothing was thrown.
```

And dropping the one-line `extraConfig` fix on its own:

```
[ERROR] desiredNodePoolIdDoesNotLeakIntoClusterState:650 expected: <null> but was: <default-pool>
```

Thanks for catching this one. Greptile's framing did point the other way and I followed it, so the
proto citation was what settled it.

hectorvent:
Fixed, and thank you for the write-up. I re-read the proto rather than my own note, and the rule is if anything more explicit than I made it: `desired_node_pool_autoscaling` spells out the carve-out you inferred, "If there is only one pool in the cluster and desired_node_pool_id is not provided then the change applies to that single node pool", so your three branches are the documented behaviour rather than a reasonable reading of it.

Resolving before assigning is the right call for the reason you give, and asserting the pools are untouched after a rejection is a better test than asserting it threw.

On splitting the `extraConfig` fix: keep it here. It is one line in the same code path, it has its own test, and it was firing on every Terraform cluster update, so it belongs with the finding that surfaced it.

No blockers from my side. The `current_node_version` minimum semantics is rightly #177.

hectorvent (approved):
Approving. Four rounds, and each one you closed the ask and found an adjacent real bug while doing it. Thank you.

PR #97, opened 2026-07-28
fix(gcs): preserve encoded names and preconditions

aniketshukla1:
## Summary

- Preserve literal URI-encoded GCS object names by avoiding a second decode of JAX-RS path parameters.
- Apply generation and metageneration preconditions inside a shared per-object lock for uploads, object updates, and deletes.
- Release idle lock entries and allocate unique object generations under concurrent writes.
- Add regression coverage for encoded object names, conditional put/patch/delete races, and lock cleanup.

Related issue: no dedicated open issue; this fixes a GCS compatibility regression identified during review.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

JAX-RS path parameters are already decoded once. Decoding them again turns a literal encoded sequence such as `%2F` into a path separator. This change preserves the object name consistently across JSON metadata, JSON download, and XML download endpoints. It also makes object preconditions atomic with the mutation they guard.

## Checklist

- [x] `./mvnw test` passes locally on JDK 25
- [x] New or updated automated tests added


aniketshukla1:
@hectorvent please review again, fixed the failing issue.
Thanks

hectorvent:
Thanks @aniketshukla1, 
please resolve conflicts when you have a chance.

aniketshukla1:
@hectorvent done. Please check. Thanks.

hectorvent:
Thanks for clearing the conflicts earlier, and sorry for the slow turnaround on my side. #120 merged today and has conflicted this branch again, but please hold off on that, see below.

**(blocking)** The resumable path is still the one gap. Session start checks preconditions before any lock, `ResumableUpload` does not carry them, and nothing re-applies them at finalize, so two competing resumable uploads with `ifGenerationMatch=0` can both succeed. Same race this PR closes for media and multipart ([generations and preconditions](https://cloud.google.com/storage/docs/generations-preconditions)).

Before you spend another round though: #93 merged on the encoded names side, and #94 now covers preconditions across every write path including resumable finalize, with SDK compat tests. Let me work out what is genuinely left here that those two do not cover, and come back to you rather than asking you to keep rebasing. Thank you for your patience on this one.

hectorvent:
Thank you again for your work here, and for your patience across three rounds. I promised to work out what remains of this PR now that #93 and #94 have merged, and the honest answer is: nothing is left. Main now checks preconditions under a per object lock on every write path, including the resumable finalize gap we discussed ([GcsService.java](https://github.com/floci-io/floci-gcp/blob/1384ebc/src/main/java/io/floci/gcp/services/gcs/GcsService.java#L242)), the delete route accepts all four precondition parameters, and generations are allocated with the same monotonic formula this PR introduced ([GcsService.java#L1124](https://github.com/floci-io/floci-gcp/blob/1384ebc/src/main/java/io/floci/gcp/services/gcs/GcsService.java#L1124)). The encoded name handling landed with #93. Your design direction is exactly where the codebase ended up; the later PRs simply raced ahead while we kept asking you to rebase, which was our fault, not yours. Rather than request a rebase that would produce an empty diff, I am closing this as superseded. I would be glad to see a future PR from you.

PR #101, opened 2026-07-30
fix(gcs): enforce the bucket CORS configuration on preflight and object reads

### LINKED ISSUE #98: [BUG] GCS: bucket CORS not applied to path-style object GET/HEAD (XML API), only to preflight
## Service

GCS

## GCP API / Method

XML API object download — `GET /{bucket}/{object}` (the URL form `blob.public_url` produces).
Compare with the JSON API `objects.get` — `GET /storage/v1/b/{bucket}/o/{object}?alt=media`, which
behaves correctly.

## Expected behavior

Real GCS applies a bucket's CORS configuration to the object response on the XML API too, so a
browser read from an allowed origin gets `Access-Control-Allow-Origin` on the actual `GET`, not just
on the preflight.

## Actual behavior

With bucket CORS set to `origin: ["http://allowed.example"]`, `method: ["GET"]`,
`responseHeader: ["Content-Type"]`, `maxAgeSeconds: 3600`:

`OPTIONS /cors-repro/hello.txt` (Origin: `http://allowed.example`) — headers present, but a fixed set
rather than the bucket's configured method/header lists:

```
HTTP/1.1 200 OK
Access-Control-Allow-Headers: Content-Type, Authorization, X-Goog-*
Access-Control-Allow-Methods: GET, POST, PUT, DELETE, OPTIONS, HEAD
Access-Control-Allow-Origin: http://allowed.example
```

`GET /cors-repro/hello.txt` (Origin: `http://allowed.example`) — no CORS headers at all:

```
HTTP/1.1 200 OK
content-length: 10
Content-Type: text/plain
```

`HEAD` on the same path is likewise bare. The preflight therefore passes and the browser then blocks
the actual read.

The JSON API download path handles the same bucket config correctly, which is what pins this to the
XML API handler rather than to CORS storage:

```
GET /storage/v1/b/cors-repro/o/hello.txt?alt=media   (Origin: http://allowed.example)
HTTP/1.1 200 OK
Access-Control-Allow-Headers: Content-Type
Access-Control-Allow-Methods: GET
Access-Control-Allow-Origin: http://allowed.example
Access-Control-Expose-Headers: Content-Type
Access-Control-Max-Age: 3600
```

This is the form the Python SDK hands to a frontend: under `STORAGE_EMULATOR_HOST`,
`blob.public_url` returns `http://localhost:4588/{bucket}/{object}`, so any browser fetch of an
uploaded object is blocked.

## Reproduction

```bash
docker run -d --name floci-gcp -p 4588:4588 floci/floci-gcp:latest
export STORAGE_EMULATOR_HOST=http://localhost:4588

curl -X POST "http://localhost:4588/storage/v1/b?project=floci-local" \
  -H 'Content-Type: application/json' -d '{"name":"cors-repro"}'

curl -X PATCH "http://localhost:4588/storage/v1/b/cors-repro" \
  -d '{"cors":[{"origin":["http://allowed.example"],"method":["GET"],"responseHeader":["Content-Type"],"maxAgeSeconds":3600}]}'

curl -X POST "http://localhost:4588/upload/storage/v1/b/cors-repro/o?uploadType=media&name=hello.txt" \
  -H 'Content-Type: text/plain' -d 'hello cors'

# Preflight: CORS headers present
curl -sD - -o /dev/null -X OPTIONS "http://localhost:4588/cors-repro/hello.txt" \
  -H 'Origin: http://allowed.example' -H 'Access-Control-Request-Method: GET'

# Actual GET: no Access-Control-Allow-Origin  <-- bug
curl -sD - -o /dev/null "http://localhost:4588/cors-repro/hello.txt" \
  -H 'Origin: http://allowed.example'

# Same object over the JSON API: headers correct
curl -sD - -o /dev/null "http://localhost:4588/storage/v1/b/cors-repro/o/hello.txt?alt=media" \
  -H 'Origin: http://allowed.example'
```

## Environment

- floci-gcp version / image tag: `floci/floci-gcp:latest` → 0.5.0
  (`org.opencontainers.image.version=0.5.0`, revision `d49271a705b73bf284e78d934f09510f70e934c1`,
  built 2026-07-09, image digest
  `sha256:102db655af6facdcd1ec621e108558aafc46c817eaada63236d3acf3c5e92aaa`, aarch64)
- GCP SDK version (if applicable): reproduction above is pure `curl`; `public_url` behaviour observed
  with `google-cloud-storage` 3.12.0 (Python)
- How you're running floci-gcp: Docker (`docker compose`), emulator reached via
  `STORAGE_EMULATOR_HOST`

### LINKED ISSUE #99: [BUG] GCS: CORS responses reflect any Origin and ignore the bucket's CORS configuration
## Service

GCS

## GCP API / Method

CORS preflight `OPTIONS` on both object URL forms
(`/{bucket}/{object}` and `/storage/v1/b/{bucket}/o/{object}`), plus JSON API `objects.get`
(`GET /storage/v1/b/{bucket}/o/{object}?alt=media`).

## Expected behavior

Real GCS evaluates the request against the bucket's CORS configuration. An origin that isn't listed
gets **no** CORS headers, and the advertised `Access-Control-Allow-Methods` /
`Access-Control-Allow-Headers` contain only what the bucket actually permits. Cross-origin code that
the bucket config forbids must fail.

## Actual behavior

Bucket CORS allows exactly one origin and one method
(`origin: ["http://allowed.example"]`, `method: ["GET"]`, `responseHeader: ["Content-Type"]`), yet a
disallowed origin requesting a disallowed method is granted access, and the advertised lists are a
fixed set unrelated to the config:

```
OPTIONS /cors-repro/hello.txt            (Origin: http://evil.example, ACRM: DELETE)
HTTP/1.1 200 OK
Access-Control-Allow-Headers: Content-Type, Authorization, X-Goog-*
Access-Control-Allow-Methods: GET, POST, PUT, DELETE, OPTIONS, HEAD
Access-Control-Allow-Origin: http://evil.example

OPTIONS /storage/v1/b/cors-repro/o/hello.txt?alt=media   (Origin: http://evil.example, ACRM: DELETE)
HTTP/1.1 200 OK
Access-Control-Allow-Headers: Content-Type, Authorization, X-Goog-*
Access-Control-Allow-Methods: GET, POST, PUT, DELETE, OPTIONS, HEAD
Access-Control-Allow-Origin: http://evil.example

GET /storage/v1/b/cors-repro/o/hello.txt?alt=media       (Origin: http://evil.example)
HTTP/1.1 200 OK
Access-Control-Allow-Origin: http://evil.example
```

The same `GET` from the *allowed* origin returns the config-derived headers, so a config-aware path
does exist — it just falls back to reflecting the request origin instead of denying it:

```
GET /storage/v1/b/cors-repro/o/hello.txt?alt=media       (Origin: http://allowed.example)
HTTP/1.1 200 OK
Access-Control-Allow-Headers: Content-Type
Access-Control-Allow-Methods: GET
Access-Control-Allow-Origin: http://allowed.example
Access-Control-Expose-Headers: Content-Type
Access-Control-Max-Age: 3600
```

Because the emulator is more permissive than GCS here, cross-origin code that production would
reject appears to work locally, and the mistake only surfaces after deploying.

## Reproduction

```bash
docker run -d --name floci-gcp -p 4588:4588 floci/floci-gcp:latest
export STORAGE_EMULATOR_HOST=http://localhost:4588

curl -X POST "http://localhost:4588/storage/v1/b?project=floci-local" \
  -H 'Content-Type: application/json' -d '{"name":"cors-repro"}'

# Bucket allows only http://allowed.example + GET
curl -X PATCH "http://localhost:4588/storage/v1/b/cors-repro" \
  -d '{"cors":[{"origin":["http://allowed.example"],"method":["GET"],"responseHeader":["Content-Type"],"maxAgeSeconds":3600}]}'

curl -X POST "http://localhost:4588/upload/storage/v1/b/cors-repro/o?uploadType=media&name=hello.txt" \
  -H 'Content-Type: text/plain' -d 'hello cors'

# Disallowed origin, disallowed method -> still granted  <-- bug
curl -sD - -o /dev/null -X OPTIONS "http://localhost:4588/storage/v1/b/cors-repro/o/hello.txt?alt=media" \
  -H 'Origin: http://evil.example' -H 'Access-Control-Request-Method: DELETE'

# Disallowed origin on the actual read -> ACAO echoed back  <-- bug
curl -sD - -o /dev/null "http://localhost:4588/storage/v1/b/cors-repro/o/hello.txt?alt=media" \
  -H 'Origin: http://evil.example'

# Allowed origin, for contrast: headers derived from the bucket config
curl -sD - -o /dev/null "http://localhost:4588/storage/v1/b/cors-repro/o/hello.txt?alt=media" \
  -H 'Origin: http://allowed.example'
```

## Environment

- floci-gcp version / image tag: `floci/floci-gcp:latest` → 0.5.0
  (`org.opencontainers.image.version=0.5.0`, revision `d49271a705b73bf284e78d934f09510f70e934c1`,
  built 2026-07-09, image digest
  `sha256:102db655af6facdcd1ec621e108558aafc46c817eaada63236d3acf3c5e92aaa`, aarch64)
- GCP SDK version (if applicable): reproduction above is pure `curl`
- How you're running floci-gcp: Docker (`docker compose`), emulator reached via
  `STORAGE_EMULATOR_HOST`

ankit1324:
## Summary

Closes #99
Closes #98

`GcsCorsFilter` never actually enforced the bucket's CORS configuration:

1. **Any origin was granted access** — `Access-Control-Allow-Origin` was set unconditionally, even when no CORS rule matched the origin.
2. **Advertised lists were unrelated to the config** — with no matching rule the filter fell back to a hardcoded `GET, POST, PUT, DELETE, OPTIONS, HEAD` / `Content-Type, Authorization, X-Goog-*`.
3. **The requested method was never checked** — `Access-Control-Request-Method` was ignored, so a preflight for a method the bucket forbids (e.g. `DELETE`) passed.
4. **Path-style (XML API) reads got no bucket CORS** — `extractBucket` only recognised the JSON API prefixes, so `GET /{bucket}/{object}` (the form `blob.public_url` produces) never resolved a bucket, which is #98.
5. The response filter's `rule == null && bucket == null` guard still emitted `Access-Control-Allow-Origin` whenever the bucket existed but permitted nothing.

Net effect: cross-origin access the bucket configuration forbids was granted, so client code (and tests) exercising a CORS-denial path silently passed against the emulator.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

Verified against the documented real-GCS behaviour ([bucket CORS](https://cloud.google.com/storage/docs/using-cors), [CORS support](https://cloud.google.com/storage/docs/cross-origin)) using the reproduction from #99/#98 — bucket CORS `origin: ["http://allowed.example"]`, `method: ["GET"]`, `responseHeader: ["Content-Type"]`, `maxAgeSeconds: 3600`:

| Request | Before | After (real GCS) |
|---|---|---|
| `OPTIONS` from `http://evil.example`, ACRM `DELETE` | 200 + `ACAO: http://evil.example` + hardcoded method list | 200, **no CORS headers** |
| `OPTIONS` from allowed origin, ACRM `DELETE` | allowed | **no CORS headers** (method not permitted) |
| `OPTIONS` from allowed origin, ACRM `GET` | fixed method/header lists | `Allow-Methods: GET`, `Allow-Headers: Content-Type`, `Max-Age: 3600` |
| `GET /{bucket}/{object}` (XML API) from allowed origin | no CORS headers | `ACAO` + `Expose-Headers: Content-Type` + `Vary: Origin` |
| `GET` from disallowed origin (either URL form) | `ACAO` reflected | **no CORS headers** |

Implementation notes:

- Rule matching now requires **both** origin and method to be permitted, honouring `*` in either list; the first matching rule wins.
- No matching rule ⇒ **no CORS headers at all**, which is what makes the browser block the request rather than the emulator silently allowing it.
- Path-style bucket extraction takes the first path segment and still goes through `findBucket`, so a non-GCS path simply resolves no bucket and no headers are added — mis-extraction is self-correcting.
- Actual (non-preflight) responses now carry only `Allow-Origin`, `Vary` and `Expose-Headers`; `Allow-Methods`, `Allow-Headers` and `Max-Age` are preflight-only per the CORS spec and are no longer emitted on object reads.

## Checklist

- [x] `./mvnw test` passes locally (full suite: **444 tests, 0 failures**)
- [x] New or updated integration test added — `GcsCorsRestIntegrationTest` (8 cases): allowed preflight advertises the configured method/header/max-age, disallowed origin and disallowed method each get no CORS headers, path-style and JSON API reads from an allowed origin carry the headers, both denial paths, and a bucket without CORS configuration gets nothing. There was no CORS coverage before.

### DIFF
--- a/src/main/java/io/floci/gcp/services/gcs/GcsCorsFilter.java
+++ b/src/main/java/io/floci/gcp/services/gcs/GcsCorsFilter.java
@@ -7,6 +7,8 @@ import jakarta.ws.rs.container.ContainerRequestContext;
 import jakarta.ws.rs.container.ContainerRequestFilter;
 import jakarta.ws.rs.container.ContainerResponseContext;
 import jakarta.ws.rs.container.ContainerResponseFilter;
+import jakarta.ws.rs.container.ResourceInfo;
+import jakarta.ws.rs.core.Context;
 import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.ext.Provider;
 
@@ -18,6 +20,9 @@ import java.util.Map;
 public class GcsCorsFilter implements ContainerRequestFilter, ContainerResponseFilter {
 
     private static final String ACCESS_CONTROL_REQUEST_METHOD = "Access-Control-Request-Method";
+
+    @Context
+    ResourceInfo resourceInfo;
 
     private final GcsService gcsService;
 
@@ -114,7 +119,7 @@ public class GcsCorsFilter implements ContainerRequestFilter, ContainerResponseF
         return false;
     }
 
-    private static String extractBucket(String path) {
+    private String extractBucket(String path) {
         String[] prefixes = {
                 "/storage/v1/b/",
                 "/download/storage/v1/b/",
@@ -128,8 +133,13 @@ public class GcsCorsFilter implements ContainerRequestFilter, ContainerResponseF
             }
         }
         // XML API (path-style) URLs — `/{bucket}/{object}`, the form `blob.public_url`
-        // produces. Any non-GCS path resolves to a bucket that does not exist, so the
-        // rule lookup simply finds nothing.
+        // produces. The leading segment is only a bucket when JAX-RS routed the request
+        // to the path-style controller; otherwise an unrelated endpoint whose first
+        // segment happens to match a bucket name (e.g. a bucket called `batch` and
+        // `/batch/storage/v1`) would inherit that bucket's CORS configuration.
+        if (!isPathStyleObjectRoute()) {
+            return null;
+        }
         String rest = path.startsWith("/") ? path.substring(1) : path;
         if (rest.isEmpty()) {
             return null;
@@ -137,6 +147,11 @@ public class GcsCorsFilter implements ContainerRequestFilter, ContainerResponseF
         int slash = rest.indexOf('/');
         String candidate = slash >= 0 ? rest.substring(0, slash) : rest;
         return candidate.isEmpty() ? null : candidate;
+    }
+
+    private boolean isPathStyleObjectRoute() {
+        return resourceInfo != null
+                && resourceInfo.getResourceClass() == GcsXmlDownloadController.class;
     }
 
     @SuppressWarnings("unchecked")
--- a/src/main/java/io/floci/gcp/services/gcs/GcsCorsFilter.java
+++ b/src/main/java/io/floci/gcp/services/gcs/GcsCorsFilter.java
@@ -17,6 +17,8 @@ import java.util.Map;
 @ApplicationScoped
 public class GcsCorsFilter implements ContainerRequestFilter, ContainerResponseFilter {
 
+    private static final String ACCESS_CONTROL_REQUEST_METHOD = "Access-Control-Request-Method";
+
     private final GcsService gcsService;
 
     @Inject
@@ -34,17 +36,19 @@ public class GcsCorsFilter implements ContainerRequestFilter, ContainerResponseF
             return;
         }
         String bucket = extractBucket(req.getUriInfo().getRequestUri().getRawPath());
-        Map<String, Object> rule = matchCorsRule(bucket, origin);
+        String requestedMethod = req.getHeaderString(ACCESS_CONTROL_REQUEST_METHOD);
+        Map<String, Object> rule = matchCorsRule(bucket, origin, requestedMethod);
         Response.ResponseBuilder rb = Response.ok();
-        rb.header("Access-Control-Allow-Origin", origin);
-        rb.header("Vary", "Origin");
+        // Real GCS only answers a preflight with CORS headers when the bucket's CORS
+        // configuration actually permits the origin and the requested method. Without a
+        // matching rule the response carries no CORS headers at all, so the browser
+        // blocks the request instead of it being silently allowed.
         if (rule != null) {
+            rb.header("Access-Control-Allow-Origin", origin);
+            rb.header("Vary", "Origin");
             appendMethodsHeader(rb, rule);
             appendHeadersHeader(rb, rule);
             appendMaxAge(rb, rule);
-        } else {
-            rb.header("Access-Control-Allow-Methods", "GET, POST, PUT, DELETE, OPTIONS, HEAD");
-            rb.header("Access-Control-Allow-Headers", "Content-Type, Authorization, X-Goog-*");
         }
         req.abortWith(rb.build());
     }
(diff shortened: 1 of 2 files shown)


ankit1324:
the route collision — it reproduced exactly as described: with a bucket named `batch`, an `OPTIONS /batch/storage/v1` request picked up that bucket's CORS configuration, because the path-style fallback read the leading segment of *every* request.

Fixed in 42f61d3 by gating the fallback on the request having actually been routed to the path-style object controller (`GcsXmlDownloadController`), which JAX-RS has already resolved by the time the filter runs — so route ownership decides it rather than a name guess. The JSON API prefixes keep their existing explicit extraction. Preflights still work on path-style URLs because that controller declares its own `@OPTIONS` methods.

Also added the missing regression test (`bucketNamedLikeAnotherRouteDoesNotGovernThatRoute`): a permissive `*`/`*` CORS config on a bucket named `batch` must not put CORS headers on `/batch/storage/v1`. It fails on the previous commit and passes now.

Full suite: 445 tests, 0 failures.

hectorvent:
Thanks for chasing this down, and the reproduction in your comment is exactly the right level of detail. Gating the path-style bucket extraction on the request actually having been routed to the XML controller is a cleaner fix than I would have suggested, and it also settles the scoping question I raised last time: unrelated endpoints no longer resolve a bucket at all, rather than resolving one that happens to share a name.

I checked the one thing that could have made this silently inert, which is whether the filter runs before matching. It does not, so `ResourceInfo` is populated for both the request and the response filter and the gate works as intended.

No blockers from my side, this looks ready to me.

hectorvent (approved):

PR #114, opened 2026-08-02
fix(firestore): enforce write preconditions and transaction conflicts

### LINKED ISSUE #113: [BUG] Firestore: write preconditions and transaction conflicts are not enforced
## Service

Firestore

## GCP API / Method

`google.firestore.v1.Firestore/Commit` (also `BatchWrite`, `Write`, `BeginTransaction`, `Rollback`)

## Expected behavior

Real Firestore (and the official Firestore emulator) enforce `Write.current_document` preconditions and serialize transactions:

- `create()` on an existing document fails with `ALREADY_EXISTS`
- `update()` on a missing document fails with `NOT_FOUND`
- `Precondition.updatedAt(...)` with a stale update time fails with `FAILED_PRECONDITION`
- A transaction whose read set was modified before commit fails with `ABORTED` (the SDK then retries), so concurrent read-modify-write transactions never lose updates

## Actual behavior

All writes are applied unconditionally:

- `create()` on an existing document succeeds and overwrites it
- `update()` on a missing document succeeds and creates it
- Stale `update_time` preconditions are silently ignored
- `BeginTransaction` returns a token but `Commit` never validates it — conflicting transactions all commit, and concurrent counter increments lose updates

No error is returned in any of these cases; the commit responds `OK`.

## Reproduction

```java
// FIRESTORE_EMULATOR_HOST=localhost:4588
Firestore db = FirestoreOptions.newBuilder()
    .setHost("localhost:4588").setProjectId("p")
    .setCredentials(NoCredentials.getInstance()).build().getService();

DocumentReference doc = db.collection("c").document("d");
doc.create(Map.of("v", 1)).get();
doc.create(Map.of("v", 2)).get();   // expected: ALREADY_EXISTS — actually succeeds

db.collection("c").document("missing").update("v", 1).get();  // expected: NOT_FOUND — actually succeeds

// Transaction conflict: expected ABORTED + SDK retry; actually both commit, one update lost
db.runTransaction(tx -> {
    long v = tx.get(doc).get().getLong("v");
    doc.update("v", 100L).get();     // out-of-band write invalidates the read
    tx.update(doc, "v", v + 1);
    return null;
}).get();
```

## Environment

- floci-gcp version / image tag: `floci/floci-gcp:latest` (0.5.0), also reproduced from `main` source
- GCP SDK version: `google-cloud-firestore` (Java) — version pinned in `compatibility-tests/sdk-test-java`
- How you're running floci-gcp: Docker

---

Fix ready. Opening a PR now.

tdewitt:
## Summary

Fixes #113

The Firestore emulator applied every write unconditionally: `Write.current_document` preconditions were ignored and `Commit` never validated its transaction, so double-creates succeeded, stale preconditions passed silently, and conflicting transactions all committed (losing updates).

- fix(firestore): enforce `current_document` preconditions on all write paths — `exists=false` → `ALREADY_EXISTS`, `exists=true` → `NOT_FOUND`, stale `update_time` → `FAILED_PRECONDITION`; `CreateDocument`/`UpdateDocument`/`DeleteDocument` pass their preconditions through
- fix(firestore): transaction conflict detection — `BeginTransaction` registers server-side state, transactional reads (`GetDocument`, `BatchGetDocuments`, `RunQuery`, incl. `new_transaction`) record the returned snapshot's version, and `Commit` aborts with `ABORTED` if any read document changed; `Rollback` discards state
- fix(firestore): `Commit` and the `Write` stream are atomic per request (preconditions validated before any write lands); `BatchWrite` is non-atomic with per-write `status`, matching real Firestore
- feat(core): add `GcpException.aborted()` (gRPC `ABORTED`, HTTP 409)

Intentional deviations (documented in `docs/services/firestore.md`): query read sets cover returned documents only (no phantom-read detection), `RunAggregationQuery` ignores transactions, transaction state is in-memory with a 15-minute TTL.

Validated with the Java SDK against a running emulator: create/update precondition failures surface as SDK exceptions, and a transaction with a conflicting concurrent write is aborted and retried by the SDK exactly once with no lost update.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/io/floci/gcp/test/FirestoreTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/floci/gcp/test/FirestoreTest.java
@@ -3,6 +3,7 @@ package io.floci.gcp.test;
 import com.google.cloud.firestore.DocumentReference;
 import com.google.cloud.firestore.DocumentSnapshot;
 import com.google.cloud.firestore.Firestore;
+import com.google.cloud.firestore.Precondition;
 import com.google.cloud.firestore.Query;
 import com.google.cloud.firestore.QueryDocumentSnapshot;
 import com.google.cloud.firestore.QuerySnapshot;
@@ -18,8 +19,10 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
 class FirestoreTest {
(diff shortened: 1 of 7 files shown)


hectorvent:
Thank you, this closes a real fidelity gap and the deviations section in the docs is exactly the right way to handle the corners you chose not to cover.

**(blocking)** In `runQuery`, the new transaction id rides on the first document response. The proto is strict here: the transaction comes back in a first response of its own, with no other fields set ([firestore.proto](https://github.com/googleapis/googleapis/blob/aa87617d67/google/firestore/v1/firestore.proto#L611-L616)). Would you be open to emitting a dedicated first message? The BatchGet path has no such constraint, so that one is fine as is.

**(follow-up, separate PR)** For an `update_time` precondition on a document that does not exist, this returns `FAILED_PRECONDITION`. The proto wording ("the target document must exist and have been last updated at that time") leaves the code ambiguous, and the [Go client's live tests](https://github.com/googleapis/google-cloud-go/blob/24c59c11/firestore/integration_test.go#L834) only cover the exists path. Did you happen to measure this case against real Firestore?

Everything else checks out against the protos and the Go client's live-service tests. Nice work on the read-race test.

tdewitt:
Nice catch on the RunQueryResponse contract.  I pushed a fix that emits the new transaction id as a dedicated first response with no other fields set.

On the update_time case... I didn't test it live. FAILED_PRECONDITION was an educated guess based on the proto language. I'll see if I can find some time to test it against a real Firestore instance and will open a new PR, if I got it wrong.

hectorvent:
Thank you for turning this around so quickly. The dedicated first response is exactly the shape the proto asks for, and pinning `!hasDocument()`, `!hasReadTime()` and `!getDone()` on it is what makes it stick, that is the whole contract in three assertions.

On your open question, half of it is already answered, so you can narrow the live test. For a wrong `update_time` on a document that does exist, `FAILED_PRECONDITION` is right, and your guess was correct: Google's own live service tests assert exactly that for both delete and update ([integration_test.go](https://github.com/googleapis/google-cloud-go/blob/24c59c11/firestore/integration_test.go#L798)). There is even a `TODO` just above wondering whether the server ought to return `InvalidArgument` instead, so the corner is openly acknowledged.

The missing document case is the part I could not find covered anywhere, so that one is still worth measuring if you get to it.

No blockers from my side.

hectorvent (approved):

PR #115, opened 2026-08-02
feat(sts): add downscoped token exchange

snazy:
## Summary

Adds the STS token-exchange endpoint used by Google Java DownscopedCredentials, a deliberately limited GCS CAB parser, global Floci credential token storage, and stored impersonated-token handling for later GCS enforcement.

Implements the STS v1 form token-exchange contract needed by the Google Java auth-library downscoping flow. Verified with STS REST tests, CAB parser tests, token-store tests, disablement coverage, and compatibility-tests/sdk-test-java StsTokenExchangeTest against a running emulator.

## Type of change

- [x] New feature (`feat:`)

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/pom.xml
+++ b/compatibility-tests/sdk-test-java/pom.xml
@@ -109,6 +109,10 @@
             <groupId>com.google.cloud</groupId>
             <artifactId>google-cloud-iamcredentials</artifactId>
         </dependency>
+		<dependency>
+			<groupId>com.google.auth</groupId>
+			<artifactId>google-auth-library-oauth2-http</artifactId>
+		</dependency>
         <dependency>
             <groupId>com.google.firebase</groupId>
             <artifactId>firebase-admin</artifactId>
--- a/compatibility-tests/sdk-test-java/src/test/java/io/floci/gcp/test/StsTokenExchangeTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/floci/gcp/test/StsTokenExchangeTest.java
@@ -1,0 +1,61 @@
+package io.floci.gcp.test;
+
+import com.google.api.client.http.javanet.NetHttpTransport;
+import com.google.auth.http.HttpTransportFactory;
+import com.google.auth.oauth2.AccessToken;
+import com.google.auth.oauth2.CredentialAccessBoundary;
+import com.google.auth.oauth2.DownscopedCredentials;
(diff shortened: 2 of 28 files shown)


hectorvent:
Thank you for slicing this out of the larger credential-vending work, the layering (parser, token store, STS surface) is easy to follow and the transport-rewrite compat test is a neat trick.

**(blocking)** The docs mark `availabilityCondition` as optional, a rule without one makes the whole bucket available ([downscoping docs](https://cloud.google.com/iam/docs/downscoping-short-lived-credentials)). The parser currently rejects that shape with `invalid_grant`, so the simplest real-world CAB fails. Would you be open to treating a missing condition as an empty prefix?

**(blocking)** Since this adds a new service, could you add a short `docs/services/sts.md` page and the `FLOCI_GCP_SERVICES_STS_ENABLED` row in the environment variables reference? AGENTS.md asks for docs alongside new services.

**(follow-up, separate PR)** Real STS ties the downscoped token's expiry to the source credential ([DownscopedCredentials](https://github.com/googleapis/google-auth-library-java/blob/9ac2d43/oauth2_http/java/com/google/auth/oauth2/DownscopedCredentials.java#L161-L171)). When the subject token is a stored floci impersonated token, inheriting its expiry instead of a flat 3600s would match that.

The wire contract itself checks out line for line against `StsRequestHandler`. Nice work.

snazy:
Had to rebase due to conflicts in the docs files.
Want me to squash all commits?

hectorvent:
Thank you for the follow up, and apologies for the slow reply to your question.

Yes please, squashing is exactly what this needs, and it clears the red check too: the Conventional Commits workflow lints every commit in the range rather than the PR title, and the second commit is titled `review`.

Both blocking points are addressed. A missing `availabilityCondition` now maps to whole bucket access, and the `docs/services/sts.md` page plus the environment variable row cover the docs ask.

**(follow-up, separate PR)** The expiry is closer but not quite there. [`DownscopedCredentials`](https://github.com/googleapis/google-auth-library-java/blob/9ac2d43/oauth2_http/java/com/google/auth/oauth2/DownscopedCredentials.java#L161-L171) puts it as "the downscoped token's expiration time will always match the source credential expiration", so a source token with more than an hour left should carry its own expiry rather than being capped at 3600. The clamp handles the shorter case, which is the half that matters most in practice.

Otherwise just the squash.

hectorvent (approved):

PR #127, opened 2026-08-17
fix(gcs): correct resumable upload chunk handling

### LINKED ISSUE #126: [BUG] GCS resumable uploads: session Location drops the port, and intermediate chunks declaring the total size get 400
## Service

GCS

## GCP API / Method

`objects.insert` with `uploadType=resumable`, and the session `PUT` requests.

## Expected behavior

1. The session `Location` is a URL the client can reach. Real GCS returns an absolute URL on the canonical host.
2. An intermediate chunk that declares the total size, e.g. `Content-Range: bytes 0-262143/524288`, gets `308 Resume Incomplete` plus a `Range` header. This is the documented chunked-upload flow ([Performing resumable uploads](https://cloud.google.com/storage/docs/performing-resumable-uploads)).

Both verified against live GCS (2026-08-17): the `Location` came back absolute on `https://storage.googleapis.com/upload/...`, and the exact intermediate chunk the emulator rejects below got `308` with `Range: bytes=0-262143`.

### A. Session `Location` loses the port when the `Host` header has none

`GcsUploadController.requestBaseUrl` returns `"http://" + Host` verbatim. The Rust SDK computes its own `Host` header and strips the port ([gax `host.rs`](https://github.com/googleapis/google-cloud-rust/blob/main/src/gax-internal/src/host.rs)), so it sends `Host: localhost` and the emulator answers:

```
Location: http://localhost/upload/storage/v1/b/.../o?uploadType=resumable&upload_id=...
```

Every chunk `PUT` then targets port 80. The SDK retries the connection failure, so `write_object(...).send_buffered()` hangs forever instead of erroring. Go, Python, Node and Java are unaffected because their HTTP stacks keep the port in `Host` (wire-verified). The fix should fall back to the actual listen
port when the header has none.

### B. Intermediate chunks that declare the total size get 400

`resumablePut` finalizes any data chunk that declares a total, instead of returning 308 while bytes are still missing:

```
PUT (session URL)  Content-Range: bytes 0-262143/524288
HTTP/1.1 400 Bad Request
{"error":{"code":400,"message":"Content-Range total size does not match uploaded bytes: 524288",...}}
```

Affected clients, each verified against the emulator with a 9 MiB payload and 256 KiB chunks:

| SDK | Intermediate chunk `Content-Range` | Result |
|---|---|---|
| Rust | total on every chunk once the payload exceeds the 8 MiB buffer | fails |
| Python | total on every chunk when the size is known | fails for uploads over 8 MiB (smaller ones take the multipart path) |
| Node | `/*` unless `metadata.contentLength` is set | fails only with explicit `contentLength` |
| Java, Go | total only on the final request | pass |

## Reproduction

```bash
docker rm -f floci-gcp 2>/dev/null
docker run -d --name floci-gcp -p 4588:4588 floci/floci-gcp:latest
curl -s -X POST "http://localhost:4588/storage/v1/b?project=floci-local" \
  -H 'Content-Type: application/json' -d '{"name":"b"}'

# Bug A: a portless Host is what the Rust SDK sends
curl -sD- -o /dev/null -X POST \
  "http://localhost:4588/upload/storage/v1/b/b/o?uploadType=resumable&name=x.bin" \
  -H 'Host: localhost' -H 'Content-Type: application/json' -d '{}'
# Location: http://localhost/upload/...   <- port 4588 is gone

# Bug B
SESSION_URL=$(curl -sD- -o /dev/null -X POST \
  "http://localhost:4588/upload/storage/v1/b/b/o?uploadType=resumable&name=y.bin" \
  -H 'Content-Type: application/json' -d '{}' \
  | tr -d '\r' | awk 'tolower($1)=="location:" {print $2}')
head -c 262144 /dev/zero > chunk1.bin
curl -sD- -X PUT "$SESSION_URL" \
  -H 'Content-Range: bytes 0-262143/524288' --data-binary @chunk1.bin
# expected: 308 + "Range: bytes=0-262143"   actual: the 400 above
```

`compatibility-tests/sdk-test-rust/tests/gcs.rs` has `resumable_upload_round_trips` marked `#[ignore]` for bug A (`cargo test resumable_upload -- --ignored` fails after its internal 60 s timeout). The fix should re-enable it. Raising its payload above 8 MiB also exercises bug B.

## Environment

- floci-gcp version / image tag: `floci/floci-gcp:0.6.6`
- GCP SDK version (if applicable): Rust `google-cloud-storage` 1.17.0 (`gax-internal` 0.7.16); cross-checked Python 3.13.1, Node 7.22.0, Java 2.47.0, Go `google.golang.org/api` v0.274.0
- How you're running floci-gcp: Docker

king-407:
## Summary

Closes #126.

This fixes two GCS resumable upload compatibility issues:

- Resumable upload session `Location` URLs now keep a reachable port when the incoming `Host` header does not include one.
- Intermediate resumable upload chunks with a known total size now return `308 Resume Incomplete` with a `Range` header until the declared total size is reached.

Previously, a portless `Host` header such as `Host: localhost` caused the session `Location` to be returned as `http://localhost/...`, which made clients target port 80 instead of the emulator port.

Previously, an intermediate chunk such as `Content-Range: bytes 0-262143/524288` was treated as final because the total size was present, causing the upload to fail with `400`.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

The issue report says this behavior was verified against live GCS on 2026-08-17:

- Live GCS returns an absolute resumable session URL on `https://storage.googleapis.com/upload/...`.
- Live GCS returns `308 Resume Incomplete` with `Range` for intermediate chunks that declare the total size.

This PR updates the emulator behavior to match that flow.

Local verification:

```bash
./mvnw -q -Dtest=GcsUploadMetadataRestIntegrationTest test
./mvnw -q test
git diff --check
```

I did not run the SDK compatibility suite locally.

## Checklist

### DIFF
--- a/src/main/java/io/floci/gcp/services/gcs/GcsService.java
+++ b/src/main/java/io/floci/gcp/services/gcs/GcsService.java
@@ -790,33 +790,42 @@ public class GcsService {
         }
         String uploadId = UUID.randomUUID().toString();
         resumableUploads.put(uploadId, new ResumableUpload(bucket, objectName, contentType,
-                customerEncryption.metadata(), metadata, preconditions, new byte[0]));
+                customerEncryption.metadata(), metadata, preconditions, new byte[0], null));
         LOG.debugf("startResumableUpload uploadId=%s", uploadId);
         return uploadId;
     }
 
     public GcsObjectMeta completeResumableUpload(String uploadId, byte[] data, String baseUrl) {
         LOG.debugf("completeResumableUpload uploadId=%s size=%d", uploadId, data.length);
-        ResumableUpload upload = resumableUploads.remove(uploadId);
+        ResumableUpload upload = resumableUploads.get(uploadId);
         if (upload == null) {
             LOG.warnf("completeResumableUpload failed: upload not found uploadId=%s", uploadId);
             throw GcpException.notFound("Resumable upload not found: " + uploadId);
         }
-        return putObject(upload.bucket(), upload.objectName(), upload.contentType(), data,
+        byte[] combined = appendChunk(upload, upload.data().length, data);
+        validateResumableTotalSize(upload, (long) combined.length);
+        resumableUploads.remove(uploadId);
+        return putObject(upload.bucket(), upload.objectName(), upload.contentType(), combined,
                 GcsCustomerEncryption.fromMetadata(upload.customerEncryption()), upload.metadata(),
                 upload.preconditions(), baseUrl);
     }
 
     public long appendResumableUpload(String uploadId, long start, byte[] data) {
+        return appendResumableUpload(uploadId, start, data, null);
+    }
+
+    public long appendResumableUpload(String uploadId, long start, byte[] data, Long totalSize) {
         ResumableUpload upload = resumableUploads.get(uploadId);
         if (upload == null) {
             LOG.warnf("appendResumableUpload failed: upload not found uploadId=%s", uploadId);
             throw GcpException.notFound("Resumable upload not found: " + uploadId);
         }
+        validateResumableTotalSize(upload, totalSize);
         byte[] combined = appendChunk(upload, start, data);
         resumableUploads.put(uploadId, new ResumableUpload(
                 upload.bucket(), upload.objectName(), upload.contentType(), upload.customerEncryption(),
-                upload.metadata(), upload.preconditions(), combined));
+                upload.metadata(), upload.preconditions(), combined,
+                upload.totalSize() != null ? upload.totalSize() : totalSize));
         return combined.length - 1L;
     }
 
@@ -835,6 +844,7 @@ public class GcsService {
             LOG.warnf("completeBufferedResumableUpload failed: upload not found uploadId=%s", uploadId);
             throw GcpException.notFound("Resumable upload not found: " + uploadId);
         }
+        validateResumableTotalSize(upload, totalSize);
         byte[] data = upload.data();
         if (data.length != totalSize) {
             throw GcpException.invalidArgument(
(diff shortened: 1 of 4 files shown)


hectorvent:
Thank you for the fix, and welcome! I verified both changes against the resumable upload protocol: an intermediate chunk that declares the total size must get `308 Resume Incomplete` with a `Range` header ([Performing resumable uploads](https://cloud.google.com/storage/docs/performing-resumable-uploads)), and the Java SDK consumes exactly the `Range: bytes=0-N` acknowledgement you emit ([JsonResumableSessionPutTask](https://github.com/googleapis/google-cloud-java/blob/e5e5f141677/java-storage/google-cloud-storage/src/main/java/com/google/cloud/storage/JsonResumableSessionPutTask.java#L126-L137)). The portless `Host` fallback to the configured port is the right call for the emulator, and exoego's live captures in #126 back both behaviors. Nice touch retaining the declared total across chunks and validating it on every completion path, the two regression tests for that are exactly right. CI is green across all suites, and I ran the full unit suite locally as well (662 tests, all green).

**(follow-up, separate PR)** A Python SDK compatibility case uploading a payload over 8 MiB would pin the chunked flow end to end, since that client declares the total on every chunk. Happy to file it as a tracking issue instead if you prefer.

No blockers from my side.

hectorvent (approved):

king-407:
Thank you for reviewing and verifying it!

I’m glad the behavior matches the resumable upload protocol. I agree that a Python SDK compatibility test for payloads over 8 MiB would be a useful follow-up. I’m happy to work on that separately if you prefer, or I can wait for a tracking issue.

PR #131, opened 2026-08-20
feat(pubsub): add IAM policy support

### LINKED ISSUE #130: [BUG] Pub/Sub: getIamPolicy on an existing topic reports "Topic not found"
## Service

Pub/Sub

## GCP API / Method

`projects.topics.getIamPolicy` / `projects.topics.setIamPolicy` (and the same for subscriptions and snapshots) — REST and the `google.iam.v1.IAMPolicy` gRPC service

## Expected behavior

Against real GCP, `GET /v1/projects/{p}/topics/{t}:getIamPolicy` on an existing topic with no policy returns an empty policy (`{"etag": "ACAB"}`), and `setIamPolicy` followed by `getIamPolicy` returns the bindings that were set. Over gRPC, the Pub/Sub SDKs (`TopicAdminClient.getIamPolicy`, Go `topic.IAM().Policy(ctx)`) call the standalone `google.iam.v1.IAMPolicy` service, since Pub/Sub declares IAM as a service-config mixin rather than on `Publisher`/`Subscriber`.

## Actual behavior

The `:getIamPolicy` suffix is not routed. The request matches the plain topic GET route (JAX-RS `[^/]+` template matches `my-topic:getIamPolicy`), which looks up a topic literally named `my-topic:getIamPolicy` and returns:

```json
{"error":{"code":404,"message":"Topic not found: projects/p/topics/t:getIamPolicy","status":"NOT_FOUND"}}
```

— for a topic that exists and that `GetTopic` returns successfully. The misleading error makes users conclude their topic is missing. `setIamPolicy` fails similarly, and the gRPC `google.iam.v1.IAMPolicy` service is not served at all (`UNIMPLEMENTED`).

This blocks any workload that provisions Pub/Sub access via IAM: Terraform/OpenTofu `google_pubsub_topic_iam_member`, a control plane granting `roles/pubsub.publisher` on a topic, or any test asserting a grant was issued.

## Reproduction

```bash
export PUBSUB_EMULATOR_HOST=localhost:4588

curl -X PUT localhost:4588/v1/projects/p/topics/t          # 200 — topic created
curl localhost:4588/v1/projects/p/topics/t                 # 200 — topic exists
curl localhost:4588/v1/projects/p/topics/t:getIamPolicy    # 404 "Topic not found" ← bug
```

## Environment

- floci-gcp version / image tag: 0.7.0
- GCP SDK version (if applicable): reproduced with Go `cloud.google.com/go/pubsub` v1.50.1 and Java `libraries-bom` 26.85.1; wire mappings verified against `googleapis/google/pubsub/v1/pubsub_v1.yaml`
- How you're running floci-gcp: `java -jar` (also reproduces under `mvn quarkus:dev`)

I have a fix ready (control-plane fidelity only — policies stored and returned, never enforced, consistent with the emulator's auth-bypass invariant) and will open a PR referencing this issue.

tdewitt:
## Summary

Closes #130

Adds the missing IAM policy surface for Pub/Sub — `getIamPolicy` / `setIamPolicy` / `testIamPermissions` on topics, subscriptions, and snapshots, over both protocols. Fixes the misleading failure where `GET :getIamPolicy` on an existing topic returned `"Topic not found"` because the suffix fell into the plain topic route.

**Control-plane fidelity only.** Policies are stored and returned, never enforced — no request is denied because of a stored policy, no condition/CEL evaluation, and `testIamPermissions` echoes the requested permissions for existing resources (empty set for missing ones — see below). This matches the emulator's accept-all-requests invariant; `docs/services/pubsub.md` documents the boundary so nobody builds an authorization test on top of it.

### How

- **gRPC** — Pub/Sub declares IAM as a service-config mixin, not on `Publisher`/`Subscriber`, so this adds a shared `google.iam.v1.IAMPolicy` implementation (`IamPolicyGrpcController`) that dispatches on resource name. One binding serves every mixin-style service on the single port; Pub/Sub is the first registrant. Uses the pre-compiled `grpc-google-iam-v1` stub (BOM-managed, same version line as the `proto-google-iam-v1` already on the classpath).
- **REST** — explicit `:getIamPolicy` (GET) / `:setIamPolicy` / `:testIamPermissions` (POST) routes on `PubSubRestController`, verbs verified against `googleapis/google/pubsub/v1/pubsub_v1.yaml`. More-literal templates win over `/topics/{topic}` — the same JAX-RS mechanism the existing `:publish` route relies on.
- **Storage** — delegates to the shared `IamService` policy store (the pattern Cloud Run already uses), with three additions:
  - **Resource existence checks** via an opt-in resolver seam: policy calls on a topic/subscription/snapshot that doesn't exist fail `NOT_FOUND`; services that haven't opted in keep today's permissive behavior. `testIamPermissions` runs the same existence check but fails open for a missing resource with an empty permission set rather than `NOT_FOUND`, per the service config's documented behavior.
  - **Etag rotation + conflict detection**: etags rotate on every write; a `setIamPolicy` carrying a stale etag fails `ABORTED`/409 (concurrency control matching real GCP, not enforcement). Previously every policy carried the constant `"ACAB"` forever, which silently lost updates in read-modify-write flows — exactly what `google_pubsub_topic_iam_member` does. Unset policies still read back with GCP's well-known empty-policy etag `ACAB`. The read-compare-write is serialized per resource with a striped lock (mirrors `GcsService.objectLock`).
  - **`deletePolicy`**: deleting a topic/subscription/snapshot clears its policy, so recreating the same name doesn't resurrect old grants.
- Bindings round-trip losslessly, including `condition` blocks — stored verbatim, never evaluated.
- Converters extracted from `CloudRunService` into a shared `IamPolicyCodec` (pure move; a codec test pins the condition round-trip).

### Intentional scope notes

- **Etag rotation is shared behavior**: Cloud Run and GCS bucket policies now rotate etags too (one store, one behavior). Nothing in the tree asserted the old constant.
- **Cloud Run had the same policy-leak-on-delete bug** — fixed here with one line in `deleteMetadata` (plus a test), rather than shipping `deletePolicy` with a single caller.
- **Schemas are excluded**: the emulator has no schema resource, so schema IAM routes would return `NOT_FOUND` unconditionally and advertise a capability that doesn't exist. Noted in docs; snapshot IAM is included since snapshots are fully implemented.
- The stale "Pub/Sub is not reachable via Terraform custom endpoints" comment in both compat `provider.tf` files is corrected, since this PR (with the existing REST controller) falsifies it.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

- Wire mappings verified against the published service config (`pubsub_v1.yaml`), not assumed — `getIamPolicy` is GET; `setIamPolicy`/`testIamPermissions` are POST with body `*`.
- **Go SDK** `cloud.google.com/go/pubsub` v1.50.1: `topic.IAM()` / `sub.IAM()` round-trips, missing-resource `NotFound`, `TestPermissions` echo — new subtests in `sdk-test-go`, run green against a live emulator.
- **Java SDK** `libraries-bom` 26.85.1: `TopicAdminClient`/`SubscriptionAdminClient` IAM round-trip + `NotFoundException` — new tests in `sdk-test-java` (19/19 green live).
- **Terraform/OpenTofu**: Pub/Sub onboarded to both IaC suites (`pubsub_custom_endpoint`, `google_pubsub_topic`, `google_pubsub_subscription`, two `google_pubsub_topic_iam_member`s on one topic plus a subscription member, bats spot checks). Verified live with OpenTofu 1.12.5 / `hashicorp/google` ~>7.36: full apply succeeds, **both topic members survive the read-modify-write**, re-plan converges with zero drift, destroy is clean. The two-members-on-one-topic case is the regression guard for the frozen-etag lost-update.
- Limitation stated per AGENTS.md: the Docker-based CI suite flow itself was not run locally; the Go/Java suites and the OpenTofu apply were executed directly against a locally built emulator instead. The `compat-terraform` HCL was validated with `tofu validate` (terraform binary not available locally); it is identical to the opentofu suite that was applied end-to-end.

## Checklist

- [x] `./mvnw test` passes locally (718 tests)
- [x] New or updated integration test added (`PubSubRestIntegrationTest` — includes the regression test for the `"Topic not found"` symptom; unit coverage for round-trip, independence, project scoping, restart survival, delete-clears-policy, condition round-trip, etag rotation/conflict/blind-write)

### DIFF
--- a/compatibility-tests/compat-opentofu/main.tf
+++ b/compatibility-tests/compat-opentofu/main.tf
@@ -161,9 +161,53 @@ resource "google_kms_crypto_key" "compat" {
   purpose  = "ENCRYPT_DECRYPT"
 }
 
+# ── Pub/Sub ───────────────────────────────────────────────────────────────────
+resource "google_pubsub_topic" "compat" {
+  name = "floci-compat-topic-tofu"
+
+  labels = {
+    env = "compat-test"
+  }
+}
+
+resource "google_pubsub_subscription" "compat" {
+  name  = "floci-compat-sub-tofu"
+  topic = google_pubsub_topic.compat.id
+
+  ack_deadline_seconds = 20
+}
+
+# Two members on one topic exercise the getIamPolicy → merge → setIamPolicy
+# read-modify-write flow; both grants must survive.
+resource "google_pubsub_topic_iam_member" "publisher" {
+  topic  = google_pubsub_topic.compat.name
+  role   = "roles/pubsub.publisher"
+  member = "serviceAccount:${google_service_account.compat.email}"
+}
+
+resource "google_pubsub_topic_iam_member" "viewer" {
+  topic  = google_pubsub_topic.compat.name
+  role   = "roles/pubsub.viewer"
+  member = "user:compat-viewer@example.com"
+}
+
+resource "google_pubsub_subscription_iam_member" "subscriber" {
+  subscription = google_pubsub_subscription.compat.name
+  role         = "roles/pubsub.subscriber"
+  member       = "serviceAccount:${google_service_account.compat.email}"
+}
+
 # ── Outputs ───────────────────────────────────────────────────────────────────
 output "key_ring_name" {
   value = google_kms_key_ring.compat.name
+}
+
+output "pubsub_topic_name" {
+  value = google_pubsub_topic.compat.name
+}
+
+output "pubsub_subscription_name" {
+  value = google_pubsub_subscription.compat.name
 }
 
 output "crypto_key_name" {
--- a/compatibility-tests/compat-opentofu/provider.tf
+++ b/compatibility-tests/compat-opentofu/provider.tf
@@ -46,8 +46,8 @@ variable "cloud_run_replace_token" {
 # floci-gcp ignores auth headers unconditionally).
 #
 # Custom endpoints redirect each service API to the local emulator.
-# Services that only expose gRPC (Pub/Sub, Firestore, Datastore) are not
-# reachable via OpenTofu custom endpoints — they need REST transcoding.
+# Services that only expose gRPC (Firestore, Datastore) are not reachable via
+# OpenTofu custom endpoints — they need REST transcoding.
 provider "google" {
   project = var.project
   region  = var.region
@@ -62,6 +62,7 @@ provider "google" {
   cloud_run_v2_custom_endpoint   = "${var.endpoint}/v2/"
   sql_custom_endpoint            = "${var.endpoint}/sql/v1beta4/"
   kms_custom_endpoint            = "${var.endpoint}/v1/"
+  pubsub_custom_endpoint         = "${var.endpoint}/v1/"
 
   # Service Usage + the Cloud Resource Manager v1 project lookup that
   # google_project_service performs on every read.
--- a/compatibility-tests/compat-opentofu/test/opentofu.bats
+++ b/compatibility-tests/compat-opentofu/test/opentofu.bats
@@ -299,7 +299,41 @@ setup() {
 
 # ── State Integrity ───────────────────────────────────────────────────────────
 
+# ── Pub/Sub Spot Checks ───────────────────────────────────────────────────────
+
+@test "OpenTofu: Pub/Sub topic created" {
+    run gcp_curl "${FLOCI_ENDPOINT}/v1/projects/${FLOCI_PROJECT}/topics/floci-compat-topic-tofu"
+    assert_success
+    assert_output --partial 'floci-compat-topic-tofu'
+}
+
+@test "OpenTofu: Pub/Sub subscription created" {
+    run gcp_curl "${FLOCI_ENDPOINT}/v1/projects/${FLOCI_PROJECT}/subscriptions/floci-compat-sub-tofu"
+    assert_success
+    assert_output --partial 'floci-compat-sub-tofu'
+}
+
+# Both google_pubsub_topic_iam_member grants must survive the provider's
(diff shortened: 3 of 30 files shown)


hectorvent:
Thank you, this is a really well put together PR, and the etag rotation fixing the lost update in `google_pubsub_topic_iam_member` is a nice catch. Routes and verbs match the Pub/Sub service config, and the 409 on a stale etag is exactly what the Terraform provider's read-modify-write loop expects.

**(blocking)** One parity point on `testIamPermissions`. The service config says that when the resource does not exist it "will return an empty set of permissions, not a `NOT_FOUND` error" ([pubsub_v1.yaml](https://github.com/googleapis/googleapis/blob/aa87617d67/google/pubsub/v1/pubsub_v1.yaml#L33-L37)), so fail-open means an empty list rather than an echo. Would you run the existence resolver there and return `[]` on miss, keeping the echo for existing resources? Your resolver seam makes it a small change. `IamPolicyGrpcControllerTest.testIamPermissionsEchoesRequestedPermissions` and the REST `topics/any` case currently pin the echo, so they would flip to expect empty. The sentence reads ambiguously, so an easy one to miss.

That is the only blocker from my side.

tdewitt:
Solid catch.  I've been doing a lot of work on a project that's GCP heavy and floci-gcp has been saving me a ton of time, money, and effort.  I'm glad to be able to contribute.  I'm burning a lot of tokens on it, since Java isn't one of my areas of expertise and I want to make sure I'm not throwing a bunch of slop your way.

Thank you for all the work you've done!

hectorvent (approved):

PR #139, opened 2026-08-27
fix(gcs): serve the resumable offset gap as text/plain

exoego:
## Summary

Follow-up to https://github.com/floci-io/floci-gcp/pull/132#pullrequestreview-5026255524

> (follow-up, separate PR) One small parity note on the offset gap 503. Real GCS serves it as plain text, and the Java SDK classifies that exact scenario by sniffing the text/plain content type ([JsonResumableSessionPutTask.java](https://github.com/googleapis/google-cloud-java/blob/e5e5f141677/java-storage/google-cloud-storage/src/main/java/com/google/cloud/storage/JsonResumableSessionPutTask.java#L217-L231)), failing fast instead of retrying. Our JSON error body means java-storage would burn its retry budget on it. Your message text already matches the real service word for word, so serving just this error as text/plain would complete the picture.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

Real GCS returns a chunk whose offset sits past the received bytes with a 503 carrying a plain text body. The Java SDK sniffs that content type in JsonResumableSessionPutTask and fails fast, so the JSON error shape made java-storage retry an error that never clears.

Checked against live GCS. The content type is "text/plain; charset=utf-8" and the message carries a double space after "Invalid request.", which puts the body at 138 bytes.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/io/floci/gcp/test/GcsResumableOffsetGapRawTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/floci/gcp/test/GcsResumableOffsetGapRawTest.java
@@ -1,0 +1,83 @@
+package io.floci.gcp.test;
+
+import com.google.cloud.storage.BucketInfo;
+import com.google.cloud.storage.Storage;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.nio.charset.StandardCharsets;
+import java.util.Locale;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * A chunk whose offset sits past the bytes the backend holds is a client-side data loss, not a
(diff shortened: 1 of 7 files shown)


hectorvent:
Merged, and thank you for picking this up, and for going further than the note asked by capturing it from the live service. Apologies that this lands after the merge rather than before it.

I rechecked the classification path and it lines up exactly: `JsonResumableSessionPutTask` gates SCENARIO_5 on [four things](https://github.com/googleapis/google-cloud-java/blob/e5e5f141677/java-storage/google-cloud-storage/src/main/java/com/google/cloud/storage/JsonResumableSessionPutTask.java#L223-L230), the status, the `text/plain` prefix, a non zero `Content-Length`, and the two substring rules, and your response satisfies all four. Asserting each separately in `GcsResumableOffsetGapRawTest` is the part I like most, since the `Content-Length` one could have regressed quietly later.

The opt in flag on `GcpException` is the right shape too: it defaults to false, `withReason` carries it through, and only the one call site uses it, so every other error keeps the JSON envelope.

One thing for later: #136 added gRPC uploads, and that path answers an offset gap with OUT_OF_RANGE while yours returns 503 text/plain. The proto pins no code for `write_offset`, so I am not yet sure which is right.

Issue #145, opened 2026-08-29
[BUG] Persistent storage has no single-writer guard: a second process silently destroys acknowledged writes


tdewitt:
Every persistent `StorageBackend` serializes its **entire in-memory map** to a temp file and renames it over the shared file. Nothing locks the data directory, so a second emulator process starts without complaint and each write is a whole-map overwrite of the other process's state. Writes that returned `200` are silently and permanently lost.

This is not specific to one service. The write lives in the three generic backends — `PersistentStorage.java:139`, `HybridStorage.java:152`, `WalStorage.java:194`, each doing `writeValue(tempFile, store)` — so it affects all 22 services that register a store (bigquery, cloudfunctions, cloudkms, cloudlogging, cloudmonitoring, cloudrun, cloudsql, credentials, datastore, eventarc, firebaseauth, firestore, gcs, gke, iam, kafka, operations, pubsub, scheduler, secretmanager, serviceusage, tasks).

## Reproduction

Two emulator processes sharing one data directory:

```
java -Dfloci-gcp.port=4611 -Dfloci-gcp.storage.mode=persistent \
     -Dfloci-gcp.storage.persistent-path=/tmp/shared -jar quarkus-run.jar &
java -Dfloci-gcp.port=4612 -Dfloci-gcp.storage.mode=persistent \
     -Dfloci-gcp.storage.persistent-path=/tmp/shared -jar quarkus-run.jar &
```

### GCS — acknowledged upload destroyed

```
pA create bucket-a                -> 200
pB create bucket-b                -> 200
pA upload bucket-a/payroll.csv    -> 200
pB upload bucket-b/other.csv      -> 200

gcs-objects.json on disk holds only bucket-b/other.csv

after stopping both and restarting a single process:
  bucket-a    = 404
  payroll.csv = 404   <-- uploaded with 200
  bucket-b    = 200
  other.csv   = 200
```

GCS object payload **bytes** are also held in a whole-map store (`gcs-object-data.json`, `StorageBackend<String, byte[]>`), so this destroys object contents, not just metadata.

### Secret Manager — completed deletion resurrected

Worse than plain loss, because it fails *toward* existence:

```
pA DELETE doomed                  -> 200, doomed correctly absent from disk
pB create bystander (unrelated)   -> 200
doomed REAPPEARS in secretmanager-secrets.json

after restart, single process:
  GET doomed = 200
  secretmanager-deletions.json = {}
```

[...]

PR #146, opened 2026-08-29
feat(iam): add project and Secret Manager policy support

### LINKED ISSUE #143: [FEAT] Support project and Secret Manager IAM policy mixins
## Problem

floci-gcp does not fully implement IAM policy mixins for Cloud Resource Manager projects and Secret Manager secrets.

Cloud Resource Manager currently exposes `projects.get` but not the project-level IAM policy methods required by standard Google Cloud clients. Secret Manager exposes placeholder IAM routes that echo a submitted policy without persisting it, so read-modify-write callers cannot retrieve the policy they set.

## Expected behavior

Implement the native GCP IAM policy mixins using the shared IAM policy infrastructure:

- Cloud Resource Manager project: `POST projects/{project}:getIamPolicy`, `setIamPolicy`, and `testIamPermissions`
- Secret Manager secret: `GET {secret}:getIamPolicy`; `POST {secret}:setIamPolicy` and `testIamPermissions`
- Persist policy bindings, return a stable empty-policy etag, rotate etags on writes, reject stale writes with `ABORTED`, and clear a secret policy when the secret is deleted.
- Project policy lookup should accept valid emulator project IDs even when no policy has been set. Secret policy calls for missing secrets must return `NOT_FOUND`.

Policies are stored for API compatibility but are not enforced as authorization boundaries by the emulator.

## Validation

Add focused unit/integration coverage for verb routing, read-modify-write etags, deletion cleanup, and `testIamPermissions`; extend Terraform and OpenTofu compatibility coverage.

tdewitt:
Closes #143. Reopens the work from #144, which I closed to rework after the review flagged that storage write failures could clear deletion intent before cleanup was durable.

## What changed since #144

The core of that review comment was that every durable backend suppressed I/O failures, so the deletion marker could be removed without any guarantee that the policy, version, and secret cleanup had reached disk. That is fixed:

- `StorageBackend.checkpoint()` replaces best-effort `flush()` on the deletion path. It throws `StorageException` when it cannot write, so a deletion that cannot be made durable surfaces as `UNAVAILABLE` instead of reporting success.
- `deleteSecret` resumes an interrupted deletion rather than returning `NOT_FOUND`.
- `listSecrets` and `getSecret` hide secrets carrying a pending deletion marker, so a half-deleted secret is never observable.
- Startup recovery logs the *cause* when a resume fails, instead of just the secret name.

## Feature

- Shared IAM policy storage, per-resource locking, etag, and cleanup behavior across services.
- Secret Manager REST IAM routes and Resource Manager project IAM mixins.
- Deletion-intent replay on startup.

## Verification

- `mvn test` — 778 tests, 0 failures.
- Persistence test covers `persistent`, `hybrid`, and `wal`, and proves state stays absent across a **second** reconstruction, not just the first.
- Crash-recovery probed against the packaged emulator, not only in unit tests: `SIGKILL` the JVM, inject a deletion marker while down, restart. Recovery completes and secret, version, IAM policy, and marker all end up absent.
- New `SecretManagerIamTest` in the Java SDK compatibility suite exercises the new IAM surface against a running emulator over gRPC, including that deleting a secret removes its policy. The IAM surface previously had no live-SDK coverage.

## Known limitation, filed separately

`checkpoint()` renames a fully written temp file, so the boundary survives process death but there is no `fsync` and it does not survive host power loss. It also assumes a single writing process per data directory. That single-writer assumption is not enforced anywhere in the storage layer and affects all 22 persisted services, not just Secret Manager — filed as #145 rather than folded in here, since a data-directory lock changes startup behavior platform-wide. The `StorageBackend.checkpoint()` javadoc now states both limits.


hectorvent:
Merged, and thank you for reworking this after #144. Apologies that this lands after the merge rather than before it.

The durability rewrite is a clear improvement, and the verb split is the part I most expected to see wrong: GET for Secret Manager's `getIamPolicy`, POST for Resource Manager's, exactly as [service.proto](https://github.com/googleapis/googleapis/blob/aa87617d67/google/cloud/secretmanager/v1/service.proto#L244) and the [Resource Manager discovery doc](https://www.googleapis.com/discovery/v1/apis/cloudresourcemanager/v1/rest) have them.

Two follow-ups for a later PR, if you are up for it.

**(follow-up)** `IamPolicyCodec` moves `etag` with `toStringUtf8` / `copyFromUtf8`, but the field is `bytes etag = 3` ([policy.proto](https://github.com/googleapis/googleapis/blob/aa87617d67/google/iam/v1/policy.proto#L157)). That round trips fine for the base64 etags the emulator generates. A gRPC client sending bytes that are not valid UTF-8 would get different bytes back, then a spurious ABORTED on the next write. Base64 would keep gRPC and REST agreeing.

**(follow-up)** The IAM routes return a bare 405 with no body, and an unexpected `permissions` or `policy` type throws ClassCastException into a plain 500, so neither carries the usual `{"error": {...}}` envelope. Worth routing both through `GcpException`.

PR #154, opened 2026-08-31
fix(firebaseauth): match Number() coercion for session cookie validDuration

ianpardal:
## Summary

Follow-up to #141, which added `projects.createSessionCookie`. Four `validDuration` inputs did not match the emulator's coercion.

The emulator computes [`Number(validDuration) || SESSION_COOKIE_MAX_VALID_DURATION`](https://github.com/firebase/firebase-tools/blob/96d0f83bd74410cb472b312bcbfab4dd03a4baeb/src/emulator/auth/operations.ts#L685), so a value falls back to the 14-day maximum **only** when it coerces to `NaN` or `0`. Anything else reaches the `[300, 1209600]` range check. Divergence ran in both directions:

| `validDuration` | Emulator | Before | After |
|---|---|---|---|
| `0.5` | `INVALID_DURATION` | 14 days | `INVALID_DURATION` |
| `true` | `INVALID_DURATION` | 14 days | `INVALID_DURATION` |
| `"0x10"` | `INVALID_DURATION` | 14 days | `INVALID_DURATION` |
| `"3600d"` | 14 days | 3600s | 14 days |
| `"\u00a03600"` (NBSP) | 3600s | 14 days | 3600s |

Causes, in order: `Number#longValue()` truncated `0.5` to `0` and took the `|| MAX` branch; `Boolean` is not a `java.lang.Number`, so `true` fell through to string parsing; `Double.parseDouble` rejects `0x` literals without a `p` exponent; and `Double.parseDouble` *accepts* Java's `d`/`f` type suffixes, which JavaScript does not.

The NBSP case came out of review: `String.trim()` only strips chars `<= U+0020`, so any of the Unicode `Zs` characters left the string unparseable. `String.strip()` is not the fix either — `Character.isWhitespace('\u00a0')` is `false` by design — so trimming follows ECMAScript `StrWhiteSpace` explicitly (`Zs` plus TAB, VT, FF, LF, CR, LS, PS, ZWNBSP, and deliberately *not* U+200B, which is `Cf` and which `Number()` leaves in place).

Coercion goes through a `JsNumber` helper implementing ECMAScript `Number()` — radix prefixes (`0x`/`0b`/`0o`, unsigned as in JS), exponent notation, `Infinity`, booleans, blank strings — and `validDuration` is carried as a `double`, so fractional values are range-checked instead of being truncated before the check.

No SDK sends any of these shapes, so this is a fidelity fix rather than a user-facing bug: `firebase-admin` always sends an integer.

### Placement

The coercion lives in its own `JsNumber` class rather than in `FirebaseAuthService`, following `FirebaseJwt` in this package and the `*Parser`/`*Codec` helpers in the other services: a package-private `final class` with a private constructor and static entry points, which is also what `AGENTS.md` asks for under "prefer package-private constructors for testability". `FirebaseAuthService` keeps only the `|| MAX` fallback and the range check, the parts specific to `createSessionCookie`.

### Structured values (arrays/objects) are deliberately not `ToPrimitive`-converted

Bare `Number([3600])` is `3600`, so it looks like this helper should return `3600` too. It should not, because the emulator never evaluates it.

`validDuration` is typed `string` in the emulator's OpenAPI schema ([`schema.ts`](https://github.com/firebase/firebase-tools/blob/96d0f83bd74410cb472b312bcbfab4dd03a4baeb/src/emulator/auth/schema.ts#L2063-L2073)), and every request body is validated against that schema by exegesis before the operation runs. The validator's only repair pass coerces `typeof value === "number"` to a string and maps enum indices to names ([`server.ts`](https://github.com/firebase/firebase-tools/blob/96d0f83bd74410cb472b312bcbfab4dd03a4baeb/src/emulator/auth/server.ts#L588-L632)); `typeof [] === "object"`, so an array matches neither repair, the ajv `type: string` error stands, and it surfaces as `Invalid JSON payload received…` / `INVALID_ARGUMENT` ([`server.ts`](https://github.com/firebase/firebase-tools/blob/96d0f83bd74410cb472b312bcbfab4dd03a4baeb/src/emulator/auth/server.ts#L248-L258)).

So for `[3600]` the emulator returns **400 `INVALID_ARGUMENT`**, never a 3600-second cookie. Teaching `JsNumber` to unwrap arrays would produce a `200` and a valid cookie where the real emulator refuses the request, which is further from the emulator, not closer.

floci-gcp has no request-schema validation layer in any service, so the closest available behavior for a structurally invalid duration is the `|| MAX` fallback, which is also what this endpoint did before this PR. Adding type rejection for `validDuration` alone would be inconsistent with the rest of the codebase; adding it project-wide is a much larger change than this PR. Happy to open that as a separate issue if maintainers want it.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

Matched against the official Firebase Auth emulator at [`firebase-tools@96d0f83`](https://github.com/firebase/firebase-tools/blob/96d0f83bd74410cb472b312bcbfab4dd03a4baeb/src/emulator/auth/operations.ts#L680-L710). Behavior was cross-checked case by case against ECMAScript `Number()` semantics, including the cases that were already correct (`null`, `""`, `"NaN"`, `"Infinity"`, `"1e3"`, whitespace-padded values, `"-0x10"`).

SDK coverage is unchanged from #141: `com.google.firebase:firebase-admin` 9.9.0 via `compatibility-tests/sdk-test-java`. gcloud CLI is not applicable — it exposes no Identity Toolkit `createSessionCookie` surface.

## Checklist

### Tests

`JsNumberTest` gives the helper direct unit coverage: 7 tests over the numeric/boolean, blank, whitespace, radix, type-suffix, decimal-literal and structured-value cases, each expectation being what a JavaScript engine returns for `Number(value)`.

At the HTTP level, `createSessionCookieCoercesDurationLikeJavaScriptNumber` covers all three outcomes — coerces-and-rejects (`0.5`, `true`, `"0x10"`, `"0b1010"`, `"1e-3"`), coerces-to-falsy-so-MAX (`"3600d"`, `"3600f"`, `"-0x10"`, `false`, `""`), and coerces-into-range (`"0x1000"` → 4096, `"3.6e3"` → 3600, `" 3600 "` → 3600). Confirmed to fail against the previous implementation and pass with this one.

Full suite:

Both the NBSP assertion and the original four-case test were confirmed to fail against the implementation each one fixes.

Full suite:

```
Tests run: 855, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

### DIFF
--- a/docs/services/firebase-auth.md
+++ b/docs/services/firebase-auth.md
@@ -40,10 +40,14 @@ the API hostname as a **path**, which floci-gcp serves directly on its single po
 - Session cookies: `createSessionCookie` re-issues a verified ID token's payload with a fresh
   `iat`/`exp` window and `iss=https://session.firebase.google.com/{project}`. `validDuration`
   is in **seconds** and must fall in `[300, 1209600]` (5 minutes – 14 days), else
-  `INVALID_DURATION`. Like the official emulator, a value that is not a non-zero number —
-  absent, `0`, or a string such as `"3600s"` — silently falls back to the 14-day maximum
-  rather than being rejected. There is no verification endpoint: the Admin SDKs verify
-  session cookies locally, and `checkRevoked` goes through `accounts:lookup`.
+  `INVALID_DURATION`. Like the official emulator, the value is first put through JavaScript
+  `Number()` coercion and then `|| MAX`, so anything coercing to `NaN` or `0` - absent,
+  `0`, `false`, `""`, or a string such as `"3600s"` silently falls back to the 14-day
+  maximum, while anything coercing to another number (`0.5`, `true`, `"0x10"`) reaches the
+  range check and is rejected there. Radix prefixes (`"0x1000"`) and exponent notation
+  (`"3.6e3"`) are read the way `Number()` reads them. There is no verification endpoint:
+  the Admin SDKs verify session cookies locally, and `checkRevoked` goes through
+  `accounts:lookup`.
 
 ## Quick Start
 
--- a/docs/services/firebase-auth.md
+++ b/docs/services/firebase-auth.md
@@ -45,9 +45,9 @@ the API hostname as a **path**, which floci-gcp serves directly on its single po
   `false`, `""`, or a string such as `"3600s"`) silently falls back to the 14-day maximum,
   while anything coercing to another number (`0.5`, `true`, `"0x10"`) reaches the range
   check and is rejected there. Radix prefixes (`"0x1000"`), exponent notation (`"3.6e3"`)
-  and ECMAScript whitespace (which includes NBSP) are read the way `Number()` reads them. There is no verification endpoint as
-  the Admin SDKs verify session cookies locally, and `checkRevoked` goes through
-  `accounts:lookup`.
+  and ECMAScript whitespace (which includes NBSP) are read the way `Number()` reads them.
+  There is no verification endpoint as the Admin SDKs verify session cookies locally, and
+  `checkRevoked` goes through `accounts:lookup`.
 
 ## Quick Start
 
--- a/docs/services/firebase-auth.md
+++ b/docs/services/firebase-auth.md
@@ -41,11 +41,11 @@ the API hostname as a **path**, which floci-gcp serves directly on its single po
   `iat`/`exp` window and `iss=https://session.firebase.google.com/{project}`. `validDuration`
   is in **seconds** and must fall in `[300, 1209600]` (5 minutes – 14 days), else
   `INVALID_DURATION`. Like the official emulator, the value is first put through JavaScript
-  `Number()` coercion and then `|| MAX`, so anything coercing to `NaN` or `0` - absent,
-  `0`, `false`, `""`, or a string such as `"3600s"` silently falls back to the 14-day
-  maximum, while anything coercing to another number (`0.5`, `true`, `"0x10"`) reaches the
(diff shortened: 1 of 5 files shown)


hectorvent (approved):
Thank you, this is careful work. Chasing `Number()` down to `StrWhiteSpace`, and deliberately leaving U+200B alone because it is `Cf` rather than a space, is the kind of detail that normally gets missed, and `JsNumberTest` reads like a spec.

I verified the contract against the emulator source. `operations.ts` computes `Number(reqBody.validDuration) || SESSION_COOKIE_MAX_VALID_DURATION` and only then range checks ([operations.ts#L685-L690](https://github.com/firebase/firebase-tools/blob/1e17461/src/emulator/auth/operations.ts#L685-L690)), so the `NaN`/`0` split is right, and carrying the value as a `double` so `0.5` reaches the range check is the correct reading.

**(follow-up, separate PR)** One nuance in your summary table: for `true` the emulator rejects at schema validation rather than at the range check, since the repair pass only coerces `typeof value === "number"` ([server.ts#L613-L616](https://github.com/firebase/firebase-tools/blob/1e17461/src/emulator/auth/server.ts#L613-L616)), so the wire error there is `INVALID_ARGUMENT`. Both are 400, and your change is closer either way. That schema validation issue you offered to open sounds worth having.

No blockers from my side.

PR #155, opened 2026-08-31
fix(gcs): populate Bucket.bucket_id on gRPC bucket responses

### LINKED ISSUE #148: [BUG] GCS gRPC: Bucket.bucket_id is never populated, so BucketInfo.getGeneratedId() is empty
## Service
GCS (Cloud Storage)

## GCP API / Method
`google.storage.v2.Storage` — `GetBucket`, `CreateBucket`, `UpdateBucket`, `ListBuckets` (any RPC returning a `Bucket`)

## Expected behavior
`Bucket.bucket_id` is an output-only field carrying the user-chosen part of the bucket name:

> Output only. The user-chosen part of the bucket name. The `{bucket}` portion of the `name` field. For globally unique buckets, this is equal to the `bucket name` of other Cloud Storage APIs. Example: `pub`.

[storage.proto#L2579](https://github.com/googleapis/googleapis/blob/aa87617d67/google/storage/v2/storage.proto#L2579)

The Java SDK consumes it directly and unconditionally when converting a gRPC `Bucket` into `BucketInfo`:

```java
to.setGeneratedId(from.getBucketId());
```

[GrpcConversions.java#L326](https://github.com/googleapis/google-cloud-java/blob/e5e5f141677/java-storage/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java#L326)

So `BucketInfo.getGeneratedId()` should return the bucket name for a gRPC client, matching what a REST client sees in the `id` field.

## Actual behavior
`GcsGrpcMapper.toProto(GcsBucket)` populates `name`, `project`, `metageneration`, `location`, `storage_class`, labels, versioning and timestamps, but never sets `bucket_id`. Because proto3 omits defaults, the field is absent on the wire and `BucketInfo.getGeneratedId()` comes back empty for gRPC clients while REST clients get a value (`GcsBucket` carries an `id` property).

Source: [`GcsGrpcMapper.java`](https://github.com/floci-io/floci-gcp/blob/main/src/main/java/io/floci/gcp/services/gcs/GcsGrpcMapper.java#L51)

## Reproduction

```java
Storage storage = StorageOptions.grpc()
        .setHost("http://localhost:4588")
        .setProjectId("test-project")
        .build()
        .getService();

storage.create(BucketInfo.of("my-bucket"));
Bucket b = storage.get("my-bucket");

// expected: "my-bucket"
// actual:   "" (empty)
System.out.println(b.getGeneratedId());
```

## Suggested fix
Set `bucket_id` to the bare bucket name in `GcsGrpcMapper.toProto(GcsBucket)`, alongside the existing `name` (`projects/_/buckets/{bucket}`). Note the field is `OUTPUT_ONLY`, so it must be emitted on responses and ignored on requests.

## Environment
- floci-gcp version: `main` @ 071393a (introduced with #136, Cloud Storage gRPC v2)
- GCP SDK: `google-cloud-storage` (Java), gRPC transport
- Reference clones: googleapis @ `aa87617d67`, google-cloud-java @ `e5e5f141677`

avison9:
## Summary

`Bucket.bucket_id` was never populated on any gRPC response, so `BucketInfo.getGeneratedId()` came back empty for every gRPC client, while REST clients reading the same bucket from the same store got a correct value. Closes #148.

`bucket_id` is `OUTPUT_ONLY` and carries the `{bucket}` portion of `name` ([storage.proto#L2579](https://github.com/googleapis/googleapis/blob/aa87617d67/google/storage/v2/storage.proto#L2579)). The Java SDK consumes it unconditionally when building a `BucketInfo`:

```java
// GrpcConversions.java:326
to.setGeneratedId(from.getBucketId());
```

Because proto3 omits defaults, an unset field is indistinguishable on the wire from a deliberate empty string. Nothing throws, and client code that keys a cache or builds a path from `getGeneratedId()` silently gets `""` on gRPC only.

```java
Storage storage = StorageOptions.grpc()
        .setHost("http://localhost:4588")
        .setProjectId("test-project")
        .build()
        .getService();

storage.create(BucketInfo.of("my-bucket"));
Bucket b = storage.get("my-bucket");

// BEFORE:
System.out.println("[" + b.getGeneratedId() + "]");   // []           <- field absent
// AFTER:
System.out.println("[" + b.getGeneratedId() + "]");   // [my-bucket]  <- matches REST `id`
```

## Root cause

`GcsGrpcMapper.toProto(GcsBucket)` set `name` (`projects/_/buckets/{bucket}`) but never set the bare `bucket_id`, dropping half of a pair the proto defines as derived from one another.

## What changed

One line in `GcsGrpcMapper.toProto(GcsBucket)`:

```java
 Bucket.Builder value = Bucket.newBuilder()
         .setName(bucketName(stored.getName()))
+        .setBucketId(stored.getName())
         .setProject("projects/" + stored.getProjectId())
```

That mapper is the sole construction site for a response `Bucket`, reached from `GcsGrpcController` at lines 81, 94, 111 and 131 (`CreateBucket`, `GetBucket`, `ListBuckets` and `UpdateBucket`), exactly the RPCs the issue names. Fixing the mapper rather than the four call sites is both smaller and leaves no path to drift.

It is sourced from `stored.getName()` rather than the REST `id` property deliberately: `getName()` is the same value `name` is built from, so the two proto fields stay consistent by construction, whereas `id` is populated separately and could be absent on a bucket restored from older persisted state.

No request-side handling is needed, because the field is `OUTPUT_ONLY` and neither `bucketCreateFields` nor `bucketUpdateFields` reads it.

## GCP Compatibility

Verified against `google-cloud-storage` (Java) over the gRPC transport, the same client and call the issue reports. Reference sources as cited in the issue: googleapis @ `aa87617d67`, google-cloud-java @ `e5e5f141677`. The change is additive on the wire: a field that was absent becomes present with the value the SDK already expects, so nothing about stored state or configuration changes.

## How it was tested

|---|---|---|---|---|---|
| Baseline (`main`, unmodified) | 851 | 0 | 0 | 0 | BUILD SUCCESS |
| With this change | 852 | 0 | 0 | 0 | BUILD SUCCESS |

No pre-existing test changed state; the delta is the one added test.

**`GcsGrpcControllerTest.everyBucketResponseCarriesBucketId`** (new) drives all four bucket-returning RPCs and asserts `getBucketId()` on each response, so the root cause is pinned rather than just the reported `GetBucket` symptom. Confirmed to fail without the fix, with only the mapper line reverted:

```
[ERROR] GcsGrpcControllerTest.everyBucketResponseCarriesBucketId:93
        expected: <bucket-id-bucket> but was: <>
```

**`GcsGrpcTest`** (compatibility-tests/sdk-test-java) gains one assertion:

```java
assertThat(storage.get(bucketName).getGeneratedId()).isEqualTo(bucketName);
```

This is the issue's own reproduction against the real SDK, per AGENTS.md's preference for SDK-based validation. It pins what a user actually observes, through the `GrpcConversions` layer that was receiving the empty value. That module is not built by `mvn test`; it was verified locally with `mvn -f compatibility-tests/sdk-test-java/pom.xml test-compile` and will execute here, since `compatibility.yml` triggers on both `src/**` and `compatibility-tests/**`.

## Risk / rollback

Low. Additive, output-only field on bucket metadata RPCs; no public Java signature changed (`GcsGrpcMapper` is package-private and `final`); no persisted format change, so builds either side of this are interchangeable against the same data directory. Rollback is reverting the single line.

## Noted, deliberately not fixed here

`GcsBucket.getEtag()` is populated (`"CAE="`) but `toProto` likewise never emits `Bucket.etag`, so gRPC clients see an empty etag where REST clients see a value. Same method, same class of bug, but whether the emulator should return a constant etag over gRPC is a separate compatibility question, so it deserves its own issue rather than being folded into this one. Happy to open it if useful.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

---

https://claude.ai/code/session_01H2RLAuMUSgQLtuteEyxyNn

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/io/floci/gcp/test/GcsGrpcTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/floci/gcp/test/GcsGrpcTest.java
@@ -35,6 +35,7 @@ class GcsGrpcTest {
                     .build());
 
             assertThat(storage.get(bucketName).getLabels()).containsEntry("transport", "grpc");
+            assertThat(storage.get(bucketName).getGeneratedId()).isEqualTo(bucketName);
             assertThat(StreamSupport.stream(storage.list().iterateAll().spliterator(), false)
                     .map(bucket -> bucket.getName())).contains(bucketName);
             assertThat(storage.update(storage.get(bucketName).toBuilder()
--- a/src/main/java/io/floci/gcp/services/gcs/GcsGrpcMapper.java
+++ b/src/main/java/io/floci/gcp/services/gcs/GcsGrpcMapper.java
@@ -51,6 +51,7 @@ final class GcsGrpcMapper {
     static Bucket toProto(GcsBucket stored) {
         Bucket.Builder value = Bucket.newBuilder()
                 .setName(bucketName(stored.getName()))
+                .setBucketId(stored.getName())
                 .setProject("projects/" + stored.getProjectId())
                 .setMetageneration(parseLong(stored.getMetageneration()))
                 .setLocation(orEmpty(stored.getLocation()))
--- a/src/test/java/io/floci/gcp/services/gcs/GcsGrpcControllerTest.java
+++ b/src/test/java/io/floci/gcp/services/gcs/GcsGrpcControllerTest.java
@@ -8,6 +8,9 @@ import com.google.storage.v2.Bucket;
 import com.google.storage.v2.ChecksummedData;
 import com.google.storage.v2.ComposeObjectRequest;
 import com.google.storage.v2.CreateBucketRequest;
+import com.google.storage.v2.GetBucketRequest;
+import com.google.storage.v2.ListBucketsRequest;
+import com.google.storage.v2.ListBucketsResponse;
 import com.google.storage.v2.ListObjectsRequest;
 import com.google.storage.v2.ListObjectsResponse;
 import com.google.storage.v2.QueryWriteStatusRequest;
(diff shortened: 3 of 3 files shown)


hectorvent:
Thank you for this, it is a model bug report: the one line lands in exactly the right place, and sourcing `bucket_id` from `getName()` rather than the REST `id` keeps the two proto fields consistent by construction.

I re-checked the contract and it holds. `storage.proto` defines `bucket_id` as "the user-chosen part of the bucket name, the `{bucket}` portion of the `name` field" ([storage.proto#L2576-L2579](https://github.com/googleapis/googleapis/blob/aa87617d67/google/storage/v2/storage.proto#L2576-L2579)), and the Java SDK reads it unguarded where its neighbours use `ifNonNull`, so an absent field really does surface as `""` ([GrpcConversions.java#L326](https://github.com/googleapis/google-cloud-java/blob/e5e5f141677/java-storage/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java#L326)).

I ran it locally too: 852/852 green, and reverting your one line does fail `GcsGrpcTest` on the `getGeneratedId` assertion, so the SDK-level coverage pins the real thing.

That `Bucket.etag` gap you spotted looks right, so a separate issue for it would be welcome.

No blockers from my side.

PR #176, opened 2026-09-02
fix(gcs): evict abandoned resumable and streaming upload sessions

### LINKED ISSUE #150: [BUG] Abandoned resumable uploads are never evicted on either the REST or gRPC path
## Service
GCS (Cloud Storage)

## GCP API / Method
Resumable uploads on both transports: REST `POST /upload/storage/v1/b/{bucket}/o?uploadType=resumable` and gRPC `google.storage.v2.Storage.BidiWriteObject` / `WriteObject`

## Expected behavior
An upload session that is started and then abandoned should not retain its buffered bytes indefinitely. Real GCS expires incomplete resumable sessions (the upload URL is valid for a bounded window, after which the session and its accumulated data are dropped).

## Actual behavior
Both in-flight upload maps are keyed by upload id and entries are removed only on an explicit terminal event. Neither has a TTL or any size bound, so a client that starts an upload and disappears leaves the buffered bytes resident for the lifetime of the process.

REST path:
```java
private final ConcurrentHashMap<String, ResumableUpload> resumableUploads = new ConcurrentHashMap<>();
```
[`GcsService.java#L68`](https://github.com/floci-io/floci-gcp/blob/main/src/main/java/io/floci/gcp/services/gcs/GcsService.java#L68) — entries are removed on finalize ([`#L1046`](https://github.com/floci-io/floci-gcp/blob/main/src/main/java/io/floci/gcp/services/gcs/GcsService.java#L1046)) or an explicit abort, and nowhere else.

gRPC path:
```java
private final ConcurrentHashMap<String, GcsStreamingUpload> streamingUploads = new ConcurrentHashMap<>();
```
[`GcsService.java#L76`](https://github.com/floci-io/floci-gcp/blob/main/src/main/java/io/floci/gcp/services/gcs/GcsService.java#L76). #136 added a release on stream close, but it deliberately covers only the non-resumable case:

```java
private void releaseNonResumableUpload(String uploadId, boolean resumable) {
    if (uploadId != null && !resumable) {
        service.abortStreamingUpload(uploadId);
    }
}
```
[`GcsGrpcController.java#L599`](https://github.com/floci-io/floci-gcp/blob/main/src/main/java/io/floci/gcp/services/gcs/GcsGrpcController.java#L599)

That carve-out is correct on its own terms, since a resumable stream is meant to survive a disconnect and be resumed later. It just means an abandoned *resumable* stream is retained exactly like the REST case.

## Impact
Long-lived emulator instances (CI runs, `compat-docker`, a dev container left up) accumulate buffered upload bytes with no upper bound. Not a wire-protocol deviation, but a resource leak.

## Suggested fix
Give both maps a common eviction policy: record a last-touched timestamp per session and drop entries past a configurable idle TTL. Worth wiring through `EmulatorConfig` (`@ConfigMapping` with `@WithDefault`) plus main and test `application.yml`, per AGENTS.md Configuration Rules. Handling both maps together keeps the two transports consistent.

## Environment
- floci-gcp version: `main` @ 071393a
- Applies to both the REST and gRPC object write paths

avison9:
## Summary

Both in-flight upload maps in `GcsService`, `resumableUploads` (REST) and `streamingUploads` (gRPC), were only ever cleared by an explicit terminal event: a finalize, an abort, or a stream close. Neither had a TTL or a size bound, so a client that started an upload and disappeared left its buffered bytes resident for the lifetime of the process, on either transport. Closes #150.

Nothing fails loudly: no error, no log line, no wire-protocol deviation. Memory grows in proportion to abandoned uploads and never comes back, which bites long-lived instances: a `compat-docker` run, a CI container, a dev container left up.

```java
// BEFORE, 1000 sessions started and never finished:
for (int i = 0; i < 1000; i++) {
    service.startResumableUpload("bucket", "abandoned-" + i + ".bin", "application/octet-stream"
            GcsCustomerEncryption.none(), Map.of());
}
// every session, and every byte pushed into it, is retained forever.
// No way to observe it, no way to reclaim it.

// AFTER, a background reaper drops sessions idle past the configured timeout:
int reclaimed = service.evictExpiredUploadSessions(System.currentTimeMillis(), idleTimeoutMillis);
// reclaimed == 1000; the buffered bytes are released.
```

## Root cause

An upload session's lifetime was tied entirely to client cooperation: every removal path required the client to come back and say something, and nothing bounded the wait.

Worth flagging that the gRPC side is **not** an oversight in the same way. #136's carve-out is correct on its own terms:

```java
private void releaseNonResumableUpload(String uploadId, boolean resumable) {
    if (uploadId != null && !resumable) {
        service.abortStreamingUpload(uploadId);
    }
}
```

A resumable gRPC stream is *meant* to survive a disconnect and be resumed later, so releasing it on stream close would break the feature. That is precisely why the fix has to be time-based rather than close-based, and why it has to cover both maps together.

## What changed

**`GcsService.evictExpiredUploadSessions(nowMillis, idleTimeoutMillis)`** (new, package-private) sweeps both maps, drops sessions whose last write is older than the timeout, and returns the count.

Two deliberate details:

- `nowMillis` is a parameter rather than a clock read, so the sweep is testable without sleeping and without threading a `Clock` into a class whose test constructors pass `config = null`.
- Each removal takes the same monitor the matching write path holds, `uploadLock(uploadId)` for resumable chunks, the session object itself for streaming writes (what `GcsGrpcController` already synchronizes on). Without that, an active upload could be evicted mid-write.

**Last-touched tracking.** `ResumableUpload` gains a `lastTouchedMillis` component, stamped on creation and on each advancing chunk, the record is already replaced per chunk, so this is free. `GcsStreamingUpload` gains a guarded `lastTouchedMillis` updated in `append`/`finalizeWith`.

**`GcsUploadSessionReaper`** (new) is the sweeper, modelled line-for-line on the existing `ScheduleDispatcher`: `@ApplicationScoped`, single daemon `ScheduledExecutorService`, `StartupEvent`/`ShutdownEvent`, and a `sweepSafely()` that swallows throwables so one bad tick cannot kill the schedule. Keeping the same shape means the codebase has one background-task idiom, not two.

**Config**, per the AGENTS.md configuration rules:

| Key | Default |
|---|---|
| `floci-gcp.services.gcs.upload-session-idle-timeout-seconds` | `604800` (7 days) |
| `floci-gcp.services.gcs.upload-session-sweep-interval-seconds` | `3600` |

Wired into `application.yml` and documented in `docs/services/gcs.md`, `docs/configuration/environment-variables.md` and `docs/configuration/advanced/application-yml.md`. Test `application.yml` is untouched, the defaults are right there and the tests drive the sweep directly.

### On the default timeout, the one judgement call

**The default matches real GCS: a seven-day resumable session window.** Given the project's rule that emulator behaviour should match real GCP, the default deviates from GCP by exactly zero.

Being straight about the tradeoff, because it is the obvious review question: a 7-day default will not reclaim anything inside the lifetime of a CI run. What this delivers *by default* is that the leak becomes bounded and configurable rather than unbounded and invisible; operators of long-lived instances lower the timeout to reclaim sooner. The alternative was a ~1-hour default, which fixes the practical symptom out of the box at the cost of expiring sessions real GCS would still honour. Happy to flip the default if maintainers would rather optimise for the emulator's actual usage than for parity, it is a one-line change to `@WithDefault`.

Setting `upload-session-sweep-interval-seconds: 0` disables the sweeper entirely, restoring exactly the previous behaviour.

## GCP Compatibility

No wire-protocol change: no request or response shape is touched, and no new endpoint exists. The only behavioural change is that a session idle beyond the configured window stops resolving, which is what real GCS does, at the same default duration. Reference: GCS resumable session URIs are valid for one week.

## How it was tested

|---|---|---|---|---|---|
| Baseline, `main` unmodified | 852 | 0 | 0 | 0 | BUILD SUCCESS |
| With this change | 856 | 0 | 0 | 0 | BUILD SUCCESS |

Four tests added to `GcsServiceTest`:

- **`abandonedUploadSessionsAreEvictedOnceIdlePastTheTimeout`**, starts one REST and one gRPC session and asserts *both* are swept together. This is the root-cause test: it fails if either transport is missed, which is exactly how this bug arose.
- **`activeUploadSessionsSurviveTheSweep`**, the safety property; a sweep that evicts live sessions would be worse than the leak.
- **`advancingAResumableChunkRefreshesTheIdleDeadline`**, pins that a chunk actually refreshes the stamp, so "idle timeout" cannot silently degrade into "age since creation".
- **`aNonPositiveIdleTimeoutDisablesEviction`**, the off switch.

Confirmed the tests detect the pre-fix behaviour. With the sweep body neutered to `return 0`, precisely what the code did before:

```
[ERROR] GcsServiceTest.abandonedUploadSessionsAreEvictedOnceIdlePastTheTimeout:398
        both the REST and the gRPC session should be swept together ==> expected: <2> but was: <0>
```

CDI wiring is exercised by startup rather than asserted: the reaper logs `GCS upload session reaper started (sweep every 3,600s, idle timeout 604,800s)` on every Quarkus boot in the suite. Its `sweep()` is a one-line delegation to the tested method, so it carries no logic of its own worth a separate test.

## Risk / rollback

Medium, and handled deliberately: this touches upload-path data structures and adds a background thread. Mitigations, the sweep takes the same locks as the writers, a test is dedicated to *not* evicting live sessions, the sweeper has a kill switch, and the GCP-parity default changes no existing behaviour. Nothing public changed signature; rollback is reverting the commit or setting the sweep interval to `0`.

## Deliberately out of scope

A byte ceiling on buffered upload data, as a second bound alongside the TTL. The issue notes there is neither "a TTL or any size bound"; the TTL is what it asks for, and a size cap is a separate policy decision about what to do when the ceiling is hit. Worth its own issue if wanted.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

---

https://claude.ai/code/session_01H2RLAuMUSgQLtuteEyxyNn

### DIFF
--- a/docs/configuration/advanced/application-yml.md
+++ b/docs/configuration/advanced/application-yml.md
@@ -56,6 +56,8 @@ floci-gcp:
   services:
     gcs:
       enabled: true
+      upload-session-idle-timeout-seconds: 604800   # real GCS resumable session window
+      upload-session-sweep-interval-seconds: 3600   # 0 disables the sweeper
     pubsub:
       enabled: true
     firestore:
--- a/docs/configuration/environment-variables.md
+++ b/docs/configuration/environment-variables.md
@@ -73,6 +73,8 @@ Each service can be toggled independently. All are enabled by default.
 | Variable | Default | Description |
 |---|---|---|
 | `FLOCI_GCP_SERVICES_GCS_ENABLED` | `true` | Cloud Storage (GCS) |
+| `FLOCI_GCP_SERVICES_GCS_UPLOAD_SESSION_IDLE_TIMEOUT_SECONDS` | `604800` | Idle time before an unfinished resumable or streaming upload session is dropped (defaults to the real GCS seven-day session window) |
+| `FLOCI_GCP_SERVICES_GCS_UPLOAD_SESSION_SWEEP_INTERVAL_SECONDS` | `3600` | How often abandoned upload sessions are swept; `0` disables the sweeper |
 | `FLOCI_GCP_SERVICES_PUBSUB_ENABLED` | `true` | Pub/Sub |
 | `FLOCI_GCP_SERVICES_FIRESTORE_ENABLED` | `true` | Firestore |
 | `FLOCI_GCP_SERVICES_DATASTORE_ENABLED` | `true` | Datastore |
--- a/docs/services/gcs.md
+++ b/docs/services/gcs.md
@@ -11,6 +11,8 @@ floci-gcp emulates Google Cloud Storage using the real GCP wire protocols:
 | Variable | Default | Description |
 |---|---|---|
 | `FLOCI_GCP_SERVICES_GCS_ENABLED` | `true` | Enable/disable Cloud Storage |
+| `FLOCI_GCP_SERVICES_GCS_UPLOAD_SESSION_IDLE_TIMEOUT_SECONDS` | `604800` | Idle time before an unfinished resumable or streaming upload session is dropped, matching the real GCS seven-day session window. Lower it on long-lived instances to reclaim buffered bytes sooner |
+| `FLOCI_GCP_SERVICES_GCS_UPLOAD_SESSION_SWEEP_INTERVAL_SECONDS` | `3600` | How often abandoned upload sessions are swept; `0` disables the sweeper |
 | `FLOCI_GCP_BASE_URL` | `http://localhost:4588` | Base URL embedded in object URLs and pre-signed URLs |
 
 ## Emulator Variable
--- a/src/main/java/io/floci/gcp/config/EmulatorConfig.java
+++ b/src/main/java/io/floci/gcp/config/EmulatorConfig.java
@@ -232,6 +232,18 @@ public interface EmulatorConfig {
     interface GcsServiceConfig {
         @WithDefault("true")
         boolean enabled();
+
+        /**
+         * Idle time after which an unfinished resumable or streaming upload session is dropped
+         * along with its buffered bytes. Defaults to the real GCS resumable session window of
+         * seven days; lower it on long-lived instances to reclaim abandoned uploads sooner.
+         */
+        @WithDefault("604800")
+        long uploadSessionIdleTimeoutSeconds();
+
+        /** Interval between sweeps for expired upload sessions. Zero or less disables the sweeper. */
+        @WithDefault("3600")
+        long uploadSessionSweepIntervalSeconds();
     }
 
     interface PubSubServiceConfig {
--- a/src/main/java/io/floci/gcp/services/gcs/GcsService.java
+++ b/src/main/java/io/floci/gcp/services/gcs/GcsService.java
@@ -945,6 +945,48 @@ public class GcsService {
         return streamingUploads.size();
     }
 
+    // Both upload maps are keyed by upload id and, before this, were only ever cleared by an
+    // explicit terminal event: a finalize, an abort, or (for non-resumable gRPC streams) a
+    // stream close. A client that starts an upload and disappears left its buffered bytes
+    // resident for the lifetime of the process, on either transport. Sweeping the two together
+    // keeps REST and gRPC from diverging again.
+    //
+    // `nowMillis` is a parameter rather than a clock read so the sweep is directly testable,
+    // and the per-session locks below are the same monitors the write paths hold, so a session
+    // cannot be written and evicted concurrently. Returns the number of sessions dropped.
+    int evictExpiredUploadSessions(long nowMillis, long idleTimeoutMillis) {
+        if (idleTimeoutMillis <= 0) {
+            return 0;
+        }
+        long cutoff = nowMillis - idleTimeoutMillis;
+        int evicted = 0;
+
+        for (String uploadId : List.copyOf(resumableUploads.keySet())) {
+            synchronized (uploadLock(uploadId)) {
+                ResumableUpload upload = resumableUploads.get(uploadId);
+                if (upload != null && upload.lastTouchedMillis() < cutoff) {
+                    resumableUploads.remove(uploadId);
+                    evicted++;
+                }
+            }
+        }
+
+        for (Map.Entry<String, GcsStreamingUpload> entry : streamingUploads.entrySet()) {
+            GcsStreamingUpload upload = entry.getValue();
+            synchronized (upload) {
+                if (upload.lastTouchedMillis() < cutoff) {
+                    streamingUploads.remove(entry.getKey(), upload);
+                    evicted++;
+                }
+            }
+        }
+
+        if (evicted > 0) {
+            LOG.debugf("Evicted %d idle upload session(s) idle longer than %dms", evicted, idleTimeoutMillis);
+        }
+        return evicted;
+    }
+
     public GcsObjectMeta finalizeStreamingUpload(String uploadId, String baseUrl) {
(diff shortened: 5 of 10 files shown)


hectorvent:
Thank you, this is a careful fix and the writeup made it quick to verify. The seven day default matches the documented lifetime ("A session URI expires after one week"), and past that GCS answers 404 ("if it's been more than a week since the upload was initiated", [resumable uploads](https://cloud.google.com/storage/docs/resumable-uploads)), which is what a swept session now returns.

On Greptile's remaining point: `queryWriteStatus` and both `onCompleted` handlers look the session up by id each time (`src/main/java/io/floci/gcp/services/gcs/GcsGrpcController.java:370`, `:459`, `:564`), so a swept id fails with NOT_FOUND. Only `onNext` holds a reference across the lookup, and your `markEvicted` guard covers that. Nothing outstanding there.

**(blocking)** #170 landed on main this week and also extended the `ResumableUpload` record (a `systemMetadata` component), so the branch now conflicts in that record and at the two `new ResumableUpload(...)` sites in `GcsService`. Keeping both new components resolves it; no semantic overlap. Would you rebase?

That rebase is the only thing between this and merge.

avison9:
Rebased onto `8c480a2`. Both new components kept, as you suggested — no semantic overlap, so it
resolved cleanly:

```java
public record ResumableUpload(String bucket, String objectName, String contentType,
        Map<String, String> customerEncryption, Map<String, String> metadata,
        GcsObjectMeta systemMetadata, GcsObjectPreconditions preconditions,
        byte[] data, Long totalSize, long lastTouchedMillis) {}
```

`systemMetadata` stays where #170 put it and `lastTouchedMillis` goes last, so the two
`new ResumableUpload(...)` sites in `GcsService` (`:1044`, `:1086`) just take both. The two doc
blocks are folded into one javadoc, keeping your `systemMetadata` wording verbatim — that comment
is the only line in the diff that is neither side's original text. `docs/services/gcs.md`
auto-merged; the two config rows sit alongside your edits untouched.

To confirm the rebase changed nothing but the conflict, I diffed the pre- and post-rebase change
sets against their respective bases. Across all ten files the only difference is `systemMetadata`
being threaded through:

```
-    customerEncryption.metadata(), metadata, preconditions, new byte[0], null));
+    customerEncryption.metadata(), metadata, systemMetadata, preconditions, new byte[0], null));
```

Everything else is byte-identical to the reviewed commits.

|---|---|---|---|---|---|
| `upstream/main` @ `8c480a2` | 888 | 0 | 0 | 0 | BUILD SUCCESS |
| Rebased branch | 893 | 0 | 0 | 0 | BUILD SUCCESS |

Delta is exactly the five tests this PR adds. I re-measured the baseline rather than reusing the
852 from the description, since the intervening PRs moved it.

`advancingAResumableChunkRefreshesTheIdleDeadline` is the useful one here: it drives
`applyResumableChunk`'s record re-construction, which is the second conflict site, so it passing
is evidence the resolution threads both components correctly and not just that it compiles.

Thanks for chasing down the `queryWriteStatus` and `onCompleted` call sites — agreed, they all
re-look-up by id, so `markEvicted` covering `onNext` is the whole of it. I've left that as is
rather than adding guards to the readers.

hectorvent (approved):
Rebase verified, thank you. I diffed the reviewed change set against the new one the same way you did and the only difference is `systemMetadata` threaded through the two `new ResumableUpload(...)` calls, plus the merged javadoc. Merge-tree is clean against main and CI is green at `26c2ecd`.

No blockers from my side.

PR #135, opened 2026-08-26
feat(iam) policy evaluation stack (WIP!)

snazy:

PR #187, opened 2026-09-08
feat(iam): add CEL allow-policy evaluator

snazy:
## Summary

Adds the disabled-by-default IAM allow-policy domain model, a restricted Nessie CEL-Java Conditions profile, principal resolution, role expansion, and pure policy evaluation. The CEL environment exposes only the Version 1 declarations and runtime overloads, with unsupported or failed conditions non-granting.

## Type of change

- [x] New feature (`feat:`)

## GCP Compatibility

Models the supported Cloud Storage IAM Conditions subset without introducing a general authentication system or changing the default no-auth emulator behavior. Conditional bindings require Uniform Bucket-Level Access.

## Follow-up

Bucket-policy APIs and enforcement are introduced by later commits. ACL authorization, project-policy inheritance, custom roles, full UBLA lifecycle behavior, and CEL profile expansion (including `extract` and regex) remain out of scope.

## Validation

Focused CEL, evaluator, normalizer, principal-resolver, configuration, and UBLA tests cover the new pure evaluation surface.


snazy:
Full IAM/CEL change set (commit-stack) in #135

snazy:
@greptile-apps please re-review and resolve your comments if the findings are resolved

snazy:
@greptile-apps please re-review and resolve your comments if the findings are resolved

snazy:
@greptile-apps please re-review and resolve your comments if the findings are resolved

snazy:
@greptile-apps please re-review and resolve your comments if the findings are resolved

snazy:
@greptile-apps please resolve your comments for the resolved findings.

snazy:
@greptile-apps I think the "whole-message IAM update semantics findings" finding is resolved.

The requested replacement behavior is not the default protobuf `FieldMask` contract. When a message field is the final path component, the supplied message is merged into the existing message, preserving omitted sibling fields. Google’s Storage testbench implements `UpdateBucket` using `FieldMask.MergeMessage` in the same way:

- [Protobuf FieldMask update semantics](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/FieldMask)
- [Google Storage testbench UpdateBucket implementation](https://github.com/googleapis/storage-testbench/blob/20d12d89a21d7981e0820b0a86d02990b3021716/testbench/grpc_server.py#L403-L457)

The concrete `enabled=false` example no longer applies to the current head. `mergeUniformBucketLevelAccess` writes the requested Boolean value unconditionally, so a present UBLA submessage can disable a stored `true` value:

https://github.com/floci-io/floci-gcp/blob/8b3d7024cf2e579402b370c2efa8b6c26077c005/src/main/java/io/floci/gcp/services/gcs/GcsGrpcMapper.java#L218-L224

The covered behaviors are now:

- `iam_config` and `iam_config.uniform_bucket_level_access` merge their supplied message while preserving omitted siblings.
- Leaf masks such as `iam_config.public_access_prevention` can explicitly clear individual fields.
- A wildcard mask replaces the complete configuration and clears omitted IAM configuration.

These cases are covered here:

https://github.com/floci-io/floci-gcp/blob/8b3d7024cf2e579402b370c2efa8b6c26077c005/src/test/java/io/floci/gcp/services/gcs/GcsGrpcControllerTest.java#L214-L270

The REST update path is also patch/merge based, so replacing a selected gRPC message would make the protocols less consistent rather than more consistent. Could you reconsider and resolve this finding?

snazy:
@greptile-apps mind resolving your review comments then?

snazy:
@greptile-apps I think, the gRPC IAM mapping finding is resolved now. Maybe perform a fresh, full review against the latest, force-pushed head?

snazy:
Looks like Greptile's happy (despite it's [recent comment](https://github.com/floci-io/floci-gcp/pull/187#issuecomment-5597832239) versus the "[top level comment](https://github.com/floci-io/floci-gcp/pull/187#issuecomment-5597546688)")

hectorvent:
Thank you, the mask handling is careful: five separate flags, a merge that starts from stored state, and `lockedTime` stripped as server-generated. Both new fields match the proto exactly, and the rename to `merge*Message` makes the semantics much easier to follow.

Two things on the bucket IAM surface, both from the [storage v1 discovery doc](https://www.googleapis.com/discovery/v1/apis/storage/v1/rest):

1. **(blocking)** `publicAccessPrevention` takes any string. The doc and the proto both say only `inherited` and `enforced` are supported, so real GCS rejects anything else. Would you validate it?
2. **(blocking)** `bucketPolicyOnly` is never populated. The doc says it "will be populated with identical information as the uniformBucketLevelAccess field" for backward compatibility, and older clients read it.
3. **(follow-up, separate PR)** Masking a whole sub-message merges here, while `objectUpdateFields` just above replaces `metadata` wholesale. I could not settle the upstream rule locally.

Greptile's two P1s from that round did not hold; the merge semantics you landed are right.

One housekeeping note: #180 merged since, so this needs a rebase in `GcsGrpcControllerTest.java`.

snazy:
Thanks for the review, I've fixed 1+2 and created #201 for the follow-up.

snazy:
@greptile-apps your comment about bucket deletion now serializing its emptiness check with many object mutations - that appears to have existed before this PR, so rather deserving a separate PR?

snazy:
@greptile-apps mind updating your confidence score and "main" comment?

snazy:
Fix for the bucket concurrency issue in #214

hectorvent:
Thank you for turning both around so quickly, and for filing #201 for the mask question rather than growing this PR.

Rechecked at the new head: `publicAccessPrevention` now rejects anything but `inherited` and `enforced` on REST insert, patch and the gRPC update path, and `bucketPolicyOnly` mirrors `uniformBucketLevelAccess` both ways, matching the [storage v1 discovery doc](https://www.googleapis.com/discovery/v1/apis/storage/v1/rest). The focused GCS and IAM tests pass locally as well.

Agreed on the bucket deletion point too, that was pre-existing, and thanks for splitting it into #214.

No blockers from my side.

hectorvent (approved):
Approving: verified at the current head, CI green.

PR #190, opened 2026-09-09
docs: refresh service and compatibility documentation

tdewitt:
## Preface

This was not requested and there's no bug open. I found myself staring at docs and hit a couple snags.  I decided to set a robot on it, to try to do a review of the docs against the current state of the code, using the current AGENTS.md as the standard.  If this isn't useful, I have no problem just closing but I wanted to see the results and let greptile give it a pass.

## Summary

- Refresh the README architecture, service inventory, and feature descriptions to match current `main`.
- Document STS, current GCS capabilities, Docker-backed data planes, and Datastore's supported transports and index behavior.
- Bring configuration, release, Docker, and compatibility-test guidance in line with the checked-in workflows and suites.
- Remove stale test counts, obsolete examples, unsupported claims, and documentation style violations.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## GCP Compatibility

Documentation only. No wire protocol behavior changes. Claims were cross-checked against the controllers, service implementations, runtime configuration, compatibility suites, and Terraform/OpenTofu provider configuration.

Validation completed:

- `./mvnw test -Dquarkus.http.test-port=0`: 971 tests passed
- `mkdocs build --strict`
- Mermaid architecture diagram render and visual inspection
- Local link, anchor, navigation, YAML, shell-snippet, suite-inventory, and provider-endpoint checks

## Checklist


tdewitt:
@hectorvent This shows as waiting on me but I'm not seeing anything that is actually still open.  Perhaps I'm missing something?

hectorvent:
Thank you for doing this unprompted, and the service pages are the strongest part. I checked the sharpest claims rather than taking them: `DatastoreController.runQuery` really has no GQL branch where the HTTP path does, gRPC `RunAggregationQuery` really returns an empty default response, and the IAM provider endpoint really is versionless in `compat-terraform/provider.tf`. The old "custom endpoints must include the API version" rule was wrong and you were right to replace it.

1. **(blocking)** In AGENTS.md, the sources-of-truth principle is right for versions and counts, but the same edit drops the protocol table, port 4588, the storage-mode list and the suite inventory. Those are stable and are what someone reads first. Would you keep them?

2. **(follow-up, separate PR)** The PR template changes are a process call, so worth splitting out for their own decision.

tdewitt:
Template changes are split out and dropped from this PR.

---

Regarding restoring the stable data, that's your call.  I'm happy to restore them if you want.

Really quickly, I'd like to explain why I cut them from AGENTS.md.  Since that file is read by agents first, it's the first opportunity for an agent to get bad info and create contradictory output.  One of the AI experts can correct me but as I understand it, if an AGENTS.md file says "a is true" and the more canonical source says "b is true," there's a high probability that it will maintain "a is true" even after reading the linked sources, because reading those is an additional tool call that only happens when an agent doesn't already think it knows a "fact."

It's true that most of those have been stable and unchanged since they were written.  But it's not 100%.

| Fact | Drift |
|---|---|
| Port 4588 | None. Unchanged since AGENTS.md landed in #31 (Jun 7). |
| Storage-mode list | None. Unchanged since #31. |
| Protocol table | Wrong Jun 7 – Aug 31. #151 corrected the gRPC row from 4 services to 11, and the base class from `GcpGrpcController` subclass to the generated `*Grpc.*ImplBase`. Stale again today: the REST JSON row lists GCS, IAM and Secret Manager, but `src/main/java` now has ~44 JAX-RS controllers. |
| Suite list | Drifts. Never picked up `sdk-test-rust`, added Aug 20 in #128; the CI matrix has changed three times, each one a suite addition. |
| CI topology, endpoint env vars, `just` recipes | None. `compat-net`, `/results` and `FLOCI_GCP_ENDPOINT` each appear in exactly one commit — the one that introduced them. |

As I said though, I have no qualms with moving them back but I wanted to have the discussion.  The two I'd still argue for leaving as pointers are the per-service protocol rows and the suite list itself, since those are the only parts with an actual drift record.  But if you want them back as they were, say the word.

hectorvent:
Thank you for splitting the template changes out, that is done: the file on the branch now matches `main` exactly.

Two things left, and the second one is ours rather than yours.

1. **(blocking)** The AGENTS.md point from my last note is still open. On this branch the file no longer mentions port 4588, the storage-mode list or the full suite inventory, and the protocol table is still gone. Those are the stable parts people read first. Would you keep them alongside the sources-of-truth principle?

2. Not your doing: we merged another PR that rewrote the suite counts in `compatibility-tests/README.md`, so this now conflicts there. That one file is the only conflict, `docs/services/gke.md` and `README.md` still merge cleanly. Sorry for the extra round trip, it is our merge ordering.

tdewitt:
I think that puts things back and adds a strong enough directive to agents that if they do anything that would change the facts in the AGENTS.md file, they're responsible for keeping things in sync.

hectorvent:
Thank you for putting the facts back, and for the drift table in your earlier reply. You were right that the protocol table and suite list had real drift history, and pinning each restored block to its owning file is a better answer than either of our starting positions.

I rechecked the restored facts against the branch rather than the commit messages: Quarkus 3.38.3, the 16 stub artifacts in `pom.xml`, the 22 plus 1 JAX-RS split with Firestore and Cloud Tasks as the only gRPC-only services, and the endpoint variables in `compatibility.yml`. All of them match, and the README conflict is gone too.

No blockers from my side. Thanks for sticking with this through four rounds.

hectorvent (approved):
Approving: restored facts re-verified against the branch, CI green.

PR #193, opened 2026-09-11
fix(pubsub): deliver messages to push subscriptions

### LINKED ISSUE #165: [BUG] Pub/Sub push subscriptions are never delivered
## Service

Pub/Sub

## GCP API / Method

`projects.subscriptions.create` (`pushConfig`), `projects.subscriptions.modifyPushConfig`, `projects.topics.publish`

## Expected behavior

A subscription created with `pushConfig.pushEndpoint` has its messages POSTed to that endpoint as they are published, in the standard push envelope:

```json
{
  "message": {
    "data": "Ym9keQ==",
    "attributes": {},
    "messageId": "1",
    "publishTime": "2026-09-01T09:00:00Z"
  },
  "subscription": "projects/p/subscriptions/s"
}
```

A 2xx response acknowledges the message; a non-2xx is retried. Where `pushConfig.oidcToken` is set, the request carries a signed `Authorization: Bearer` header.

`docs/services/pubsub.md` states that this is supported: _"floci-gcp supports push subscriptions — it delivers messages to an HTTP endpoint you configure"_, with a Java `PushConfig` example. The 0.6.0 CHANGELOG entry for the subscription-filter fix also refers to it: _"pull, streaming pull, and push delivery skip non-matching messages instead of delivering everything"_.

## Actual behavior

Nothing is ever delivered to the push endpoint. The subscription is created successfully, the publish succeeds and returns a `messageId`, and the message then sits in the subscription backlog indefinitely, retrievable only by an explicit `pull`. No error or warning is returned or logged, so a push consumer sees only silence.

Three routes into a push subscription were tested; none delivers, and two also lose part of the config:

| Route                                         | Endpoint persisted?                                             | Delivered? |
| --------------------------------------------- | --------------------------------------------------------------- | ---------- |
| REST `subscriptions.create` with `pushConfig` | Yes, but `oidcToken` is silently dropped from the stored config | No         |
| gRPC `CreateSubscription` with `pushConfig`   | No — `GetSubscription` returns an empty `pushConfig`            | No         |
| gRPC `ModifyPushConfig`                       | Yes — read back correctly                                       | No         |

Ruling out the obvious causes on our side:

- **Not a networking problem.** A `curl` from inside the emulator container to the push endpoint returns `HTTP 204` (see the reproduction below).
- **Not the request being rejected.** The server logs and acknowledges the calls, e.g. `PubSubSubscriberController modifyPushConfig subscription=projects/p/subscriptions/s`.
- **Not a slow tick.** We waited 40 seconds, polling every 5.
- **Not delivered-then-dropped.** After waiting, the message is still in the backlog unacked — a delivered message would be gone.

No `push` or `deliver` log line appears at any level while this happens.

Accepting a `pushEndpoint` and never delivering to it is worse than rejecting it: the subscription looks correctly configured and the publish succeeds, so there is no signal that the messages are going nowhere.

## Reproduction

```bash
docker run -d --rm -p 4588:4588 --name floci-repro floci/floci-gcp:latest
B=http://localhost:4588/v1/projects/p

# Any endpoint that logs POSTs, reachable from the container.
python3 -c "
from http.server import BaseHTTPRequestHandler, HTTPServer
class H(BaseHTTPRequestHandler):
    def do_POST(self): print('PUSH RECEIVED', self.path, flush=True); self.send_response(204); self.end_headers()
    def log_message(self,*a): pass
HTTPServer(('0.0.0.0',8099),H).serve_forever()" &

curl -s -X PUT $B/topics/t
curl -s -X PUT $B/subscriptions/s -H 'Content-Type: application/json' \
  -d '{"topic":"projects/p/topics/t","pushConfig":{"pushEndpoint":"http://host.docker.internal:8099/hook"},"ackDeadlineSeconds":60}'
curl -s -X POST $B/topics/t:publish -H 'Content-Type: application/json' \
  -d '{"messages":[{"data":"Ym9keQ=="}]}'
# => {"messageIds":["1"]}

# The endpoint is reachable from inside the container — prints 204:
docker exec floci-repro curl -s -o /dev/null -w '%{http_code}\n' \
  -X POST http://host.docker.internal:8099/probe

sleep 10
# The receiver logged only the /probe, never the message. It is still queued:
curl -s -X POST $B/subscriptions/s:pull -H 'Content-Type: application/json' \
  -d '{"maxMessages":10}'
```

Observed on the final `pull`:

```json
{
  "receivedMessages": [
    {
      "ackId": "7b025c67-a90d-4611-b4a6-5039884b7c2c",
      "message": { "messageId": "1", "data": "Ym9keQ==" }
    }
  ]
}
```

Real Pub/Sub POSTs the envelope to `/hook` and leaves an empty backlog.

The same result occurs over gRPC. Creating the subscription with the official SDK and reading it back shows the endpoint was not even stored:

```javascript
process.env.PUBSUB_EMULATOR_HOST = "localhost:4588";
const { PubSub } = require("@google-cloud/pubsub");
const ps = new PubSub({ projectId: "p" });

await ps.createTopic("t2");
const [sub] = await ps.topic("t2").createSubscription("s2", {
  pushConfig: { pushEndpoint: "http://host.docker.internal:8099/hook" },
  ackDeadlineSeconds: 60,
});
console.log((await sub.getMetadata())[0].pushConfig); // => null

// Setting it afterwards does persist, but still never delivers:
await sub.modifyPushConfig({
  pushEndpoint: "http://host.docker.internal:8099/hook",
});
console.log((await sub.getMetadata())[0].pushConfig);
// => { attributes: {}, pushEndpoint: 'http://host.docker.internal:8099/hook' }

await ps.topic("t2").publishMessage({ data: Buffer.from("body") });
// 35s later: no POST received, message still in the backlog.
```

## Environment

- floci-gcp version / image tag: `floci/floci-gcp:latest` — reports `floci-gcp 0.8.0` (native) on startup
- GCP SDK version (if applicable): `@google-cloud/pubsub` 5.3.0 for the gRPC routes; raw REST via `curl` otherwise
- How you're running floci-gcp (Docker / native / `mvn quarkus:dev`): Docker, via Docker Desktop on macOS 15 (arm64)

AnuragRaut08:
## Summary

Fix Pub/Sub push subscriptions so messages are delivered to the configured `pushEndpoint`.

Previously, push subscriptions could be created with a push endpoint, but published messages remained in the subscription backlog instead of being delivered to the endpoint. The gRPC subscription creation path also failed to persist `pushConfig.pushEndpoint`.

This PR:

* Preserves `pushConfig.pushEndpoint` when creating subscriptions via gRPC.
* Delivers published messages to configured push endpoints using the Pub/Sub push envelope.
* Prevents successfully push-delivered messages from remaining in the pull queue.
* Adds regression coverage for push delivery and the request payload.

Fixes #165

## Type of change

* [x] Bug fix (`fix:`)
* [ ] New feature (`feat:`)
* [ ] Breaking change (`feat!:` or `fix!:`)
* [ ] Docs / chore

## GCP Compatibility

The bug was verified against the expected Google Cloud Pub/Sub push subscription behavior: a published message is delivered as a JSON push envelope containing the message data, attributes, message ID, publish time, and subscription name, and a successful 2xx response acknowledges the message.

Regression coverage verifies the push payload and confirms that the successfully delivered message is not available through the subscription pull queue.

## Checklist

* [x] `mvn test` passes locally
* [x] New or updated integration/regression test added
* [x] Commit messages follow Conventional Commits


hectorvent:
Thank you, and the retry test standing up a real local HTTP server is a good way to prove it.

First: three of the bot's P1 comments above were written against an earlier push and no longer describe this code. The envelope does carry `orderingKey`, delivery does go through an executor rather than blocking publish, and failures do retry. Not outstanding, so please skip them.

The other two still apply, and they are really one thing:

1. **(blocking)** Neither executor is shut down anywhere, and `schedulePushRetry` reschedules unconditionally, so an unreachable endpoint retries forever, including after its subscription is deleted. Would you add lifecycle cleanup and a stop condition?

2. **(follow-up, separate PR)** The fixed one second retry is much tighter than real Pub/Sub, where `RetryPolicy.minimum_backoff` "Defaults to 10 seconds" ([pubsub.proto](https://github.com/googleapis/googleapis/blob/aa87617d67/google/pubsub/v1/pubsub.proto#L1636-L1647)).

The envelope shape matches the default `PubsubWrapper` form, so that part reads right.

AnuragRaut08:
Thanks @hectorvent — addressed the lifecycle/retry concerns, including the replacement-subscription race.

* Both push executors are now shut down with the service lifecycle.
* Scheduled retries re-check the current subscription state, so deleted, detached, or push-disabled subscriptions are not retried.
* Retries also validate the subscription creation identity, preventing an old failed delivery from being retried to a newly recreated subscription with the same name.
* Added regression coverage for deletion, detachment, and delete/recreate under the same subscription name.
* Full test suite passes: **1032 tests, 0 failures**.

I’ll keep the Pub/Sub retry backoff change separate as suggested.

hectorvent:
Thanks for turning the lifecycle and retry race around so quickly, and CI is fully green now, so the earlier hold on my side is cleared. I re-read the delivery path against the push documentation this round and found two things I should have checked in the first pass, sorry.

**(blocking)** The wrapped envelope carries each identity field twice, [`messageId` with `message_id` and `publishTime` with `publish_time`](https://cloud.google.com/pubsub/docs/push). Webhook samples in several languages read the snake_case spelling, so they would see nothing today. Would you emit both?

**(blocking)** Acknowledgement is not the whole 2xx range: the same page lists 102, 200, 201, 202 and 204 as acks and says every other code is a negative acknowledgement. A `Set.of(...)` check instead of `>= 200 && < 300` would match.

**(follow-up, separate PR)** A permanently failing endpoint reschedules every second forever; a backoff or attempt cap would be kinder to a local machine.

Everything else matches: base64 data, attributes, `orderingKey` only when set, and the full subscription name.

Issue #196, opened 2026-09-13
[BUG] GKE: UpdateCluster stores desiredMasterVersion / desiredNodeVersion aliases verbatim, so gcloud upgrade leaves the cluster on version "-"


avison9:
## Service

GKE (Kubernetes Engine)

## GCP API / Method

`google.container.v1.ClusterManager.UpdateCluster` (`PUT .../clusters/{cluster}`), `ClusterUpdate.desired_master_version` and `ClusterUpdate.desired_node_version`

## Expected behavior

Both fields document the same aliases as `UpdateMasterRequest.master_version`, with one difference in what `-` means:

- `desired_master_version` ([cluster_service.proto#L3706-L3716](https://github.com/googleapis/googleapis/blob/aa87617d67/google/container/v1/cluster_service.proto#L3706-L3716)): `latest`, `1.X`, `1.X.Y`, `1.X.Y-gke.N`, and `-` "picks the default Kubernetes version".
- `desired_node_version` ([cluster_service.proto#L3532-L3543](https://github.com/googleapis/googleapis/blob/aa87617d67/google/container/v1/cluster_service.proto#L3532-L3543)): same list, but `-` "picks the Kubernetes master version", i.e. the cluster's current control plane version.

After the call, `currentMasterVersion` / `currentNodeVersion` (and each upgraded pool's `version`) hold a real resolved version.

## Actual behavior

`GkeService.updateCluster` stores both fields verbatim (`GkeService.java` around lines 324-343 on main @ c534f6b). After `desiredMasterVersion: "-"` the cluster reports `currentMasterVersion: "-"`; after `desiredNodeVersion: "latest"` the cluster and pool report version `"latest"`.

This is the path gcloud takes. `gcloud container clusters upgrade CLUSTER --master` and `... --node-pool POOL` both call `UpdateCluster`, and when no `--cluster-version` is given gcloud sends the literal `-` ([api_adapter.py `UpdateClusterCommon`](https://github.com/twistedpair/google-cloud-sdk/blob/master/google-cloud-sdk/lib/googlecloudsdk/api_lib/container/api_adapter.py): `if not options.version: options.version = '-'`, then `ClusterUpdate(desiredMasterVersion=options.version)` / `ClusterUpdate(desiredNodeVersion=options.version, ...)`). So the most common gcloud upgrade invocation leaves the emulated cluster on version `-`. Terraform's `min_master_version` also goes through `desiredMasterVersion`, though it always sends an explicit version.

[...]

PR #197, opened 2026-09-13
fix(gke): reject master version spellings the field does not document

### LINKED ISSUE #195: [BUG] GKE: UpdateMaster resolves a bare major "1" as a version alias
## Service

GKE (Kubernetes Engine)

## GCP API / Method

`google.container.v1.ClusterManager.UpdateMaster` (`POST .../clusters/{cluster}:updateMaster`), `UpdateMasterRequest.master_version`

## Expected behavior

`master_version` documents exactly five spellings ([cluster_service.proto#L4649-L4659](https://github.com/googleapis/googleapis/blob/aa87617d67/google/container/v1/cluster_service.proto#L4649-L4659)):

- `latest`, `-`
- `1.X`, `1.X.Y` (a minor or a patch prefix of a valid version)
- `1.X.Y-gke.N` (an explicit version)

A bare major such as `"1"` is not among them, and real GKE rejects it with `400 INVALID_ARGUMENT`.

## Actual behavior

`GkeService.resolveMasterVersion` (added in #192) matches a prefix with `DEFAULT_MASTER_VERSION.startsWith(requested + ".")`, so `"1"` matches `"1.30.5-gke.1014001"` and is resolved to the advertised version instead of being rejected. Raised by @hectorvent in the #192 review as a follow-up; cosmetic while a single version is advertised, but the helper's contract should match the field's before it grows.

## Reproduction

```bash
H=http://localhost:4588/container/v1/projects/p/locations/us-central1
curl -s -X POST "$H/clusters" -H 'Content-Type: application/json' -d '{"cluster":{"name":"c"}}'
curl -s -X POST "$H/clusters/c:updateMaster" -H 'Content-Type: application/json' -d '{"masterVersion":"1"}'
# floci-gcp: 200, operation UPGRADE_MASTER, cluster now reports 1.30.5-gke.1014001
# real GKE: 400 INVALID_ARGUMENT
```

## Environment

- floci-gcp version / image tag: main @ c534f6b
- How you're running floci-gcp (Docker / native / `mvn quarkus:dev`): any

## Suggested fix

Match the documented shapes only: `latest`, `-`, `1.X`, `1.X.Y`, and an explicit version. Reject anything else that is not an explicit version with `400 INVALID_ARGUMENT`, keeping the existing rule that an explicit version not in the advertised list is stored verbatim. Pin `"1"` (rejected) and `"1.30"` / `"1.30.5"` (resolved) in `GkeServiceTest`.

- [x] Yes
- [ ] No

avison9:
## Summary

Follow-up from the #192 review. `resolveMasterVersion` matched the `1.X` / `1.X.Y` aliases with a plain `startsWith`, so a bare major such as `"1"` resolved to the advertised version simply because it is a character prefix of `"1.30.5-gke.1014001"`. `UpdateMasterRequest.master_version` documents exactly five spellings and real GKE rejects the rest. Closes #195.

```bash
# BEFORE
H=http://localhost:4588/container/v1/projects/p/locations/us-central1
curl -s -X POST "$H/clusters" -H 'Content-Type: application/json' -d '{"cluster":{"name":"c"}}'
curl -s -X POST "$H/clusters/c:updateMaster" -H 'Content-Type: application/json' -d '{"masterVersion":"1"}'
# 200, operationType UPGRADE_MASTER; cluster now reports 1.30.5-gke.1014001

# AFTER
curl -s -X POST "$H/clusters/c:updateMaster" -H 'Content-Type: application/json' -d '{"masterVersion":"1"}'
# {"error":{"code":400,"message":"Invalid master version \"1\": expected \"latest\", \"-\", \"1.X\", \"1.X.Y\" or \"1.X.Y-gke.N\"","status":"INVALID_ARGUMENT",...}}
```

### What changed

- `GkeService.resolveMasterVersion` now classifies the input before resolving: `latest` / `-` pick the advertised version; a `1.X` / `1.X.Y` alias (`^\d+\.\d+(?:\.\d+)?$`) resolves when it prefixes the advertised version and is otherwise kept as before; an explicit `1.X.Y-gke.N` passes through verbatim as before; anything else is `400 INVALID_ARGUMENT`.
- Behaviour for every documented shape is unchanged from #192; only undocumented shapes move from "silently resolved or stored" to rejected.
- `docs/services/gke.md` updated accordingly.

Independent of #196 (different lines of the same file); either can merge first.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

Spellings taken from `UpdateMasterRequest.master_version` in `cluster_service.proto` @ `aa87617d67` (L4649-L4659). Real GKE returns `400 INVALID_ARGUMENT` for a version it cannot resolve.

## Tests

- `GkeServiceTest.updateMasterRejectsAVersionShapeTheFieldDoesNotDocument` (new): `"1"`, `"1."`, `"v1.30"`, `"banana"`, `"1.30.5-gke"`, `"1.30.5-gke."` all 400 and leave the cluster and etag untouched; an explicit non-advertised `1.X.Y-gke.N` is still stored verbatim. Written first; failed on `"1"` before the change.
- `GkeUpdateMasterRestIntegrationTest.updateMasterRejectsAnUndocumentedVersionShape` (new): the 400 error shape over REST.
- Existing alias tests unchanged and passing (`1.3` vs `1.30` non-match, `-`, `latest`, prefixes).

Full suite: baseline on `main` @ `c534f6b` is 1043 run / 0 failures / 0 errors / 0 skipped; with this change 1045 / 0 / 0 / 0 (the 2 new tests, nothing else changed).

## Checklist


hectorvent:
Thank you for turning the follow-up into its own PR with a failing test first, that is exactly the right shape.

I checked the spelling list again rather than trusting my own earlier note: the field documents five ([cluster_service.proto](https://github.com/googleapis/googleapis/blob/aa87617d67/google/container/v1/cluster_service.proto#L4649-L4659)), and your two regexes plus two literals cover exactly those. I also traced the cases by hand: `1.30` still resolves, `1.3` is still kept verbatim as the non-match, and `1` is now rejected, which was the bug. `GkeServiceTest` and `GkeUpdateMasterRestIntegrationTest` run 51/0 here.

Keeping an unmatched `1.X` alias verbatim rather than rejecting it is the right call on a single version catalogue, and the javadoc saying so is appreciated.

No blockers from my side.

hectorvent (approved):
Five documented spellings confirmed verbatim against cluster_service.proto; alias cases traced by hand; tests 51/0 locally.

PR #198, opened 2026-09-13
fix(gke): resolve version aliases on UpdateCluster desired versions

avison9:
## Summary

`UpdateCluster` stored `desiredMasterVersion` and `desiredNodeVersion` verbatim. Both fields document the same aliases as `UpdateMasterRequest.master_version` (`latest`, `-`, `1.X`, `1.X.Y`), and gcloud relies on them: `gcloud container clusters upgrade C --master` and `... --node-pool P` both call `UpdateCluster`, and when no `--cluster-version` is given gcloud sends the literal `-` (`api_adapter.py`, `UpdateClusterCommon`). Against the emulator the cluster then reported `currentMasterVersion: "-"`. Closes #196.

This is the gap noted in #192, where `UpdateMaster` gained alias resolution and `UpdateCluster` was deliberately left for its own PR.

```bash
# BEFORE
H=http://localhost:4588/container/v1/projects/p/locations/us-central1
curl -s -X POST "$H/clusters" -H 'Content-Type: application/json' -d '{"cluster":{"name":"c"}}'
curl -s -X PUT "$H/clusters/c" -H 'Content-Type: application/json' -d '{"update":{"desiredMasterVersion":"-"}}'
curl -s "$H/clusters/c" | jq .currentMasterVersion
# "-"

# AFTER
curl -s "$H/clusters/c" | jq .currentMasterVersion
# "1.30.5-gke.1014001"
```

### What changed

- `desiredMasterVersion` goes through the existing `resolveMasterVersion`, so it behaves exactly like `UpdateMaster`.
- `desiredNodeVersion` goes through a new one-line `resolveNodeVersion`: identical rules, except `-` picks the cluster's current master version rather than the server default, which is the one documented difference between the two fields (`cluster_service.proto` L3542: "picks the Kubernetes master version"). The resolved value is applied to both the cluster aggregate and each targeted pool.
- Block order in `updateCluster` is unchanged: the node block still resolves its targets before mutating anything, so a rejected `desiredNodePoolId` leaves no partial state. The proto allows one `ClusterUpdate` field per request, so a node `-` reads the stored master version, not one the same request might also be changing.
- `docs/services/gke.md` updated.

Independent of #195 (different lines of the same file); either can merge first.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Breaking change (`feat!:` or `fix!:`)
- [ ] Docs / chore

## GCP Compatibility

[...]

PR #199, opened 2026-09-14
fix(gke): name the request field in invalid version errors

avison9:
## Summary

Follow-up requested in the #198 review. With #197 and #198 both merged, `masterVersion`, `desiredMasterVersion` and `desiredNodeVersion` share one resolver, so a rejected node version reported the wrong field:

```bash
# BEFORE
curl -s -X PUT "$H/clusters/c" -H 'Content-Type: application/json' -d '{"update":{"desiredNodeVersion":"banana"}}'
# {"error":{"code":400,"message":"Invalid master version \"banana\": expected ...","status":"INVALID_ARGUMENT",...}}

# AFTER
# {"error":{"code":400,"message":"Invalid desiredNodeVersion \"banana\": expected ...","status":"INVALID_ARGUMENT",...}}
```

The status was already right; only the message text changes.

### What changed

- `resolveMasterVersion` and `resolveNodeVersion` take the request field name; the three call sites pass `masterVersion`, `desiredMasterVersion` and `desiredNodeVersion`.
- No behaviour change beyond the message.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

Error status unchanged (`400 INVALID_ARGUMENT`); the message now refers to the JSON field name the client sent, as GCP error messages do.

## Tests

- `GkeServiceTest.invalidVersionErrorsNameTheFieldThatCarriedThem` (new): all three paths with `"banana"`, asserting the message starts with `Invalid <field> "banana"`. Failed first with `Invalid master version "banana"`.
- Existing message assertions were prefix-free, so nothing else changed.

Full suite: 1059 run / 0 failures / 0 errors / 0 skipped on top of `main` @ `1bd6583` (one new test).

## Checklist


hectorvent:
Thanks for picking this up straight after #198, it is exactly the follow-up I had in mind.

I checked all three callers of the shared resolver (UpdateCluster `desiredNodeVersion` and `desiredMasterVersion`, UpdateMaster `masterVersion`), and each now passes its own field name, spelled as the proto3 JSON names of [`desired_node_version`, `desired_master_version` and `master_version`](https://github.com/googleapis/googleapis/blob/aa87617d67/google/container/v1/cluster_service.proto#L3543). The status stays 400 INVALID_ARGUMENT, and the new test pins all three paths. Ran `GkeServiceTest` locally, 50/0.

No blockers from my side.

hectorvent (approved):
Approving: all three resolver call sites verified, CI green, GkeServiceTest 50/0 locally.

PR #201, opened 2026-09-14
fix(gcs): honor gRPC object metadata key masks

snazy:
## Summary

Preserve unmasked custom metadata when UpdateObject selects `metadata.<key>`, update selected keys present in the request, and remove selected keys omitted from it. A whole metadata mask and the wildcard mask continue to replace the complete map.

The mapper now carries granular metadata changes separately from ordinary object fields so the service can apply the read-modify-write atomically under its existing object lock.

## Type of change

- [x] Bug fix (fix:)
- [ ] New feature (feat:)
- [ ] Breaking change (feat!: or fix!:)
- [ ] Docs / chore

## GCP Compatibility

Matches the Cloud Storage v2 `UpdateObject` behavior implemented by Google's storage-testbench. The Go Cloud Storage client emits `metadata.<key>` masks for non-empty `ObjectAttrsToUpdate.Metadata` maps, so its compatibility test now verifies that an existing unmasked key survives.

## Checklist

- [ ] ./mvnw test passes locally. Focused GCS service, controller, and gRPC integration tests pass; the full suite was not run.

The Go compatibility suite was not compiled because this checkout lacks required go.sum entries and dependency mutation was outside this change.


hectorvent:
Thanks for this, a tidy fix with the right split between the mapper and the locked service write.

I checked the wire shape against both clients: the Go client sends one `metadata.<key>` path per key it sets ([grpc_client.go](https://github.com/googleapis/google-cloud-go/blob/24c59c11/storage/grpc_client.go#L765-L775)), and the Java client deletes a key by masking `metadata.<key>` while dropping the null value from the map ([GrpcConversions.java](https://github.com/googleapis/google-cloud-java/blob/e5e5f141677/java-storage/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java#L1017)). Your merge for present keys and removal for absent ones matches both, and `*` and a bare `metadata` path still replace the whole map as the [update_mask docs](https://github.com/googleapis/googleapis/blob/aa87617d67/google/storage/v2/storage.proto#L2033-L2042) describe. The focused gRPC tests pass locally too (24/0).

No blockers from my side.

hectorvent (approved):
Approving: re-verified against main after #187 landed, GCS suite 265/0 on the merge result.

Issue #48, opened 2026-06-13
[Service] Implement Cloud SQL for PostgreSQL (REST control plane + Docker data plane)


shirkevich:
## Summary
Implement **Cloud SQL for PostgreSQL** in phases. Start with the Cloud SQL Admin API control plane so SDKs, gcloud, and IaC tools can create/list/update/delete Cloud SQL resources, then wire those resources to real PostgreSQL containers through the existing Docker/lifecycle infrastructure.

The goal is Cloud SQL compatibility, not a simplified SQL convenience API. The control plane should expose GCP-compatible REST shapes; the data plane should speak the normal PostgreSQL wire protocol through actual Postgres instances.

## Protocol / transport
- Control plane: REST JSON Cloud SQL Admin API (`sqladmin.googleapis.com`) over floci-gcp's normal REST port.
- Primary API version: `v1beta4` first, because many Cloud SQL examples/providers historically target it; add `v1` compatibility where low-cost or required by SDK/gcloud behavior.
- Proto / resource model: Google-published Cloud SQL protos under `google/cloud/sql/{v1,v1beta4}` define resource messages, REST annotations, operations, connect settings, users, databases, and instance schemas.
- Data plane: PostgreSQL wire protocol through Docker-backed `postgres` containers. Do not introduce custom HTTP SQL endpoints.

## Endpoint / auth bypass
- Admin API base: emulate `https://sqladmin.googleapis.com`, e.g. `/sql/v1beta4/projects/{project}/instances`.
- Auth: follow existing emulator behavior and accept credentials/tokens unconditionally.
- gcloud: investigate `gcloud sql ...` endpoint override support, including `CLOUDSDK_API_ENDPOINT_OVERRIDES_SQL`, against `http://localhost:4588`.
- Java Admin SDK: generated `google-api-services-sqladmin` clients support root URL/service path override; use this for SDK compatibility tests.
- Data-plane connection details should be surfaced through instance/connect settings metadata and point to Docker-managed local Postgres endpoints.

## Authoritative references
- REST docs: https://cloud.google.com/sql/docs/postgres/admin-api/rest
- REST Discovery: https://sqladmin.googleapis.com/$discovery/rest?version=v1beta4 and https://sqladmin.googleapis.com/$discovery/rest?version=v1
- Proto: https://github.com/googleapis/googleapis/tree/master/google/cloud/sql/v1beta4
  - `cloud_sql.proto`
  - `cloud_sql_resources.proto`
[...]

PR #202, opened 2026-09-14
feat(cloudsql): support MySQL instances backed by a Docker data plane

### LINKED ISSUE #200: [FEAT] Cloud SQL: support MySQL instances (MYSQL_8_0, MYSQL_8_4) with a Docker data plane
## Service

Cloud SQL

## API Method / Feature

`sqladmin` v1 / v1beta4: `instances.insert` with `databaseVersion: MYSQL_8_0` / `MYSQL_8_4`, plus the `databases.*` and `users.*` surface on such an instance, backed by a real MySQL data plane the way PostgreSQL instances already are.

## GCP Documentation

- https://cloud.google.com/sql/docs/mysql/db-versions
- `SqlDatabaseVersion` enum: https://cloud.google.com/sql/docs/mysql/admin-api/rest/v1/SqlDatabaseVersion
- `users.insert` (MySQL users are host-qualified, default host `%`): https://cloud.google.com/sql/docs/mysql/admin-api/rest/v1/users/insert

## Why is this needed?

`instances.insert` with any `MYSQL_*` version returns `400 Only PostgreSQL Cloud SQL instances are supported`. #48 shipped Cloud SQL as PostgreSQL only and noted "MySQL/SQL Server support should be separate follow-up scope"; this is that follow-up for MySQL.

Cloud SQL for MySQL is the more common of the two engines in existing estates, and both sister emulators already serve MySQL (floci: RDS MySQL / Aurora; floci-az: Azure SQL, PostgreSQL). Tooling that provisions `google_sql_database_instance` with `database_version = "MYSQL_8_0"` has no local target today.

## Proposed scope

- `MYSQL_8_0` and `MYSQL_8_4` (and the `MYSQL_8_0_NN` minor-pinned spellings mapping to the 8.0 image); `MYSQL_5_7` left out as it is in extended support only.
- A MySQL data plane next to `CloudSqlPostgresDataPlane` (`mysql` image, port 3306, `/var/lib/mysql`, `mysqladmin ping` readiness, `mysql` client for DDL), with the shared container lifecycle and volume naming factored into one place rather than duplicated. Dispatch on `databaseVersion`.
- Engine-aware defaults: system databases (`mysql`, `information_schema`, `performance_schema`, `sys`) listed and protected from deletion the way `postgres` is; `utf8mb4` / `utf8mb4_0900_ai_ci` database defaults; host-qualified users with `%` default, which PostgreSQL rejects today; `root@%` as the built-in user; `flags.list` entries that apply to MySQL.
- `FLOCI_GCP_SERVICES_CLOUDSQL_MYSQL80_IMAGE` / `MYSQL84_IMAGE` config, docs, and a `CloudSqlAdminTest` case in `compatibility-tests/sdk-test-java` connecting over JDBC like the PostgreSQL one.
- Not in scope: SQL Server; MySQL-specific extras such as `instances.import`/`export`, replicas, or `rootPassword` rotation beyond initial provisioning.

- [x] Yes
- [ ] No

avison9:
## Summary

Adds Cloud SQL for MySQL (`MYSQL_8_0`, `MYSQL_8_0_NN`, `MYSQL_8_4`) alongside PostgreSQL, the follow-up #48 explicitly set aside ("MySQL/SQL Server support should be separate follow-up scope"). Closes #200.

Each MySQL instance runs in an official `mysql` container the way PostgreSQL instances run in `postgres`, and the Admin API surface (`instances`, `databases`, `users`, `flags`) follows the MySQL edition of the API: system schemas listed and protected, `utf8mb4` defaults, host-qualified users, and the `root@%` account the instance is provisioned with.

```bash
# BEFORE
H=http://localhost:4588/v1/projects/p
curl -s -X POST "$H/instances" -H 'Content-Type: application/json' \
  -d '{"name":"my-main","databaseVersion":"MYSQL_8_4","region":"us-central1"}'
# {"error":{"code":400,"message":"Only PostgreSQL Cloud SQL instances are supported","status":"INVALID_ARGUMENT",...}}

# AFTER
curl -s -X POST "$H/instances" -H 'Content-Type: application/json' \
  -d '{"name":"my-main","databaseVersion":"MYSQL_8_4","region":"us-central1"}'
# {"kind":"sql#operation","status":"DONE","operationType":"CREATE",...}   (~12 s: container start + init)
curl -s "$H/instances/my-main" | jq '{databaseVersion, state, ip: .ipAddresses[0]}'
# {"databaseVersion":"MYSQL_8_4","state":"RUNNABLE","ip":{"type":"PRIMARY","ipAddress":"localhost","port":45451}}
curl -s "$H/instances/my-main/databases" | jq '[.items[].name]'
# ["information_schema","mysql","performance_schema","sys"]
curl -s -X POST "$H/instances/my-main/databases" -H 'Content-Type: application/json' -d '{"name":"appdb"}' >/dev/null
curl -s -X POST "$H/instances/my-main/users" -H 'Content-Type: application/json' -d '{"name":"app","password":"secret"}' >/dev/null
curl -s "$H/instances/my-main/users" | jq '[.items[] | "\(.name)@\(.host)"]'
# ["app@%","root@%"]
mysql -h 127.0.0.1 -P 45451 -u app -psecret appdb -e "SELECT VERSION(), @@collation_database"
# 8.4.11   utf8mb4_0900_ai_ci
```

### What changed

- **`CloudSqlEngine`** (new enum): the engine facts the control plane needs, resolved from `databaseVersion`: version prefix, system databases, default charset/collation, host qualification, built-in user. `CloudSqlService` reads these instead of hard-coding PostgreSQL.
- **`CloudSqlContainerDataPlane`** (new abstract base): the container lifecycle, volume naming (including the pinned legacy prefix backfill), readiness wait and endpoint metadata lifted verbatim out of the PostgreSQL plane. Engines supply image, port, env, data directory and readiness probe. `CloudSqlPostgresDataPlane` now extends it and keeps only its `psql` logic; `CloudSqlPostgresDataPlaneTest` passes unchanged, which is the guard that the extraction preserved behaviour.
- **`CloudSqlMySqlDataPlane`** (new): `mysql` image, port 3306, `/var/lib/mysql`, `MYSQL_ROOT_PASSWORD`. Readiness is `mysqladmin ping` over TCP on purpose, because the image's entrypoint runs a socket-only temporary server during initialisation. DDL goes through the `mysql` client with `MYSQL_PWD` in the environment so the client's password warning never pollutes the error text. Users are `'name'@'host'`; grants are `ALL PRIVILEGES ON db.*` per user database; system schemas and `root` are never touched.
- **`CloudSqlEngineDataPlane`** (new): the bean the service injects; dispatches to the engine plane by `databaseVersion` on every call.
- **`CloudSqlDataPlane`** interface: user methods take a `host`, since a MySQL identity is not just a name. PostgreSQL ignores it (the service never passes one for that engine).
- **Service**: `createInstance` accepts both engines (SQL Server still `400`); `databases.list` seeds the engine's system databases and `databases.delete` refuses them; database defaults per engine; `users.*` normalise `host` per engine (`%` default on MySQL, rejected on PostgreSQL as before); `root@%` is listed as `BUILT_IN` and cannot be deleted; `flags.list` gains MySQL entries. `rootPassword` is stripped from stored/returned instances on both engines, as it is write-only in the real API.
- **Config/docs**: `FLOCI_GCP_SERVICES_CLOUDSQL_MYSQL80_IMAGE` (`mysql:8.0.46`), `MYSQL84_IMAGE` (`mysql:8.4.11`); native-image `initialize-at-run-time` entries for the two new plane classes; `docs/services/cloud-sql-postgres.md` (kept at its path to avoid breaking links; title and nav now say PostgreSQL and MySQL), env var docs, README.

### Design notes

- One behaviour worth knowing: `instances.insert` stays synchronous, and a cold MySQL data directory takes 15 to 25 s to initialise, past the 20 s read timeout the Google Java HTTP client defaults to (PostgreSQL is ready in ~3 s). Documented, and the compat fixture raises its read timeout. The real API's asynchronous model (`RUNNING` operation, `PENDING_CREATE` instance) is the proper fix for both engines and is left for a follow-up since it changes the existing PostgreSQL contract.
- `root`'s password stays `root` regardless of `rootPassword` or `users.update`, mirroring how the PostgreSQL plane keeps `postgres`/`postgres`; documented as a limitation.
- Not included: SQL Server, `MYSQL_5_7` (extended support only), import/export, replicas.
- Follow-up agreed separately: list the built-in `postgres` user on PostgreSQL instances the way `root@%` is listed here; `CloudSqlEngine.builtInUser` is where that lands.

### Review fixes (second commit)

Greptile's first pass found six real issues, each now fixed and pinned by a test: cross-engine `databaseVersion` on patch/update is `400`; only the provisioned `root@%` is protected (other hosts are ordinary accounts); `charset`/`collation` are applied in `CREATE DATABASE` after validation as bare names; the host is percent-encoded in user storage keys; `rootPassword` is stripped on `instances.patch` too; `MYSQL_8_0` / `MYSQL_8_0_NN` are matched strictly.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

Wire shapes from the SQL Admin v1 discovery document (`sql#instance`, `sql#database`, `sql#user` with `host`, `sql#flag.appliesTo`) and the MySQL edition of the docs (system schemas on a fresh instance, `root@%`, `SqlDatabaseVersion` values). Verified with `google-api-services-sqladmin` in `compatibility-tests/sdk-test-java` (`CloudSqlAdminTest.sqlAdminSdkCreatesUsableMysqlInstances`: create instance, database and user for 8.0 and 8.4, then connect over `mysql-connector-j` and run DDL/DML) against a local instance with the Docker data plane on; the PostgreSQL case in the same test passes on the refactored plane. Also exercised by hand: password rotation, grants on later databases, `users.delete` revoking access, instance deletion leaving no container or volume.

## Tests

- `CloudSqlServiceTest` (+4): MySQL defaults (system databases, `root@%`, `utf8mb4`, `rootPassword` stripped); system databases and `root` undeletable; host-qualified identities (`%` default, explicit host is a distinct user, data plane receives host on every call); unsupported engines `400`.
- `CloudSqlMySqlDataPlaneTest` (new, 6): image/env/port/mount and TCP readiness probe; `MYSQL_8_0_NN` mapping; startup failure removes container and fresh volume; the exact SQL issued for database/user/grant/drop including identifier and literal quoting; system schemas and `root` never touched; no exec without a container.
- `CloudSqlMySqlRestIntegrationTest` (new, 1): the Admin JSON shapes end to end in mock mode.
- `CloudSqlPostgresDataPlaneTest`: unchanged, passing on the extracted base.
- `CloudSqlRestIntegrationTest`: the "non-Postgres is rejected" case now uses SQL Server, since MySQL is accepted.
- Compat: `CloudSqlAdminTest` +1 (JDBC over `mysql-connector-j` 9.3.0), README counts updated.

Full suite: baseline on `main` @ `1bd6583` is 1058 run / 0 failures / 0 errors / 0 skipped; with this change 1074 / 0 / 0 / 0 (the 16 new tests, nothing else changed).

## Checklist


avison9:
Addressed in 6b7d890, one test per finding:

- **Engine change via patch/update**: rejected with 400 when `databaseVersion` would cross engines; same-engine version changes keep their existing behaviour (`instanceEngineCannotBeChangedByPatchOrUpdate`).
- **root at other hosts**: only the provisioned `root@%` is protected or skipped; `root@10.0.0.5` is an ordinary account, created on the server and deletable (`onlyTheProvisionedRootIdentityIsProtected`, plane test).
- **Collation ignored**: `CREATE DATABASE` now carries `CHARACTER SET`/`COLLATE` from the resource; both are validated as bare names before being spliced into DDL (plane test includes a refused injection attempt).
- **Key collision**: the host is percent-encoded in the user storage key, so the first `/` always ends it; PostgreSQL keys are unchanged (`hostQualifiedUserKeysCannotCollide`).
- **PATCH and rootPassword**: stripped on patch too (`patchDoesNotPersistRootPassword`).
- **Malformed 8.0 versions**: strict `MYSQL_8_0` / `MYSQL_8_0_NN` match; `MYSQL_8_00`, `MYSQL_8_0foo`, `MYSQL_8_4_1` rejected.
- **Docs**: the `operations.get` suggestion is gone; the timeout note now says only to raise the read timeout.

The branch was also rebased onto main to drop an unrelated GKE commit (#199) that had been carried in by mistake; the diff is now Cloud SQL only. Full suite 1074/0/0/0.

avison9:
Second pass addressed in 45be253:

- **root@localhost**: right, the image provisions root at localhost/127.0.0.1/::1 as well, and the plane's own 127.0.0.1 connections authenticate through those. They are now reserved alongside root@%: refused on insert and delete, skipped by the plane; root at any other host stays an ordinary account (`localRootIdentitiesAreReservedOnMysql`, plane test).
- **Charset-only requests**: the engine defaults now apply only when the request names neither charset nor collation; naming one applies and reports just that one, so latin1 is never paired with utf8mb4_0900_ai_ci (`charsetAndCollationAreDefaultedTogetherOrNotAtAll`).

Full suite 1075/0/0/0. The follow-up for the built-in postgres user on PostgreSQL instances is https://github.com/floci-io/floci-gcp/pull/204, stacked on this branch with a merge-order note.

avison9:
One more from the #204 review, addressed here in b795193: the exact `databaseVersion` is now validated in `CloudSqlEngine` rather than only when the Docker plane picks an image, so `cloudsql.mock=true` no longer accepts `MYSQL_5_7` or `POSTGRES_14` (the latter was pre-existing). Full suite 1076/0/0/0.

hectorvent:
Thank you for this, the plane extraction is clean and the greptile rounds were handled with a test per finding, which made this easy to check.

1. **(blocking)** `users.delete` refuses `root@%`, but the Terraform provider deletes exactly that user right after creating any MySQL instance ([resource_sql_database_instance.go#L1811-L1843](https://github.com/hashicorp/terraform-provider-google/blob/4de3c23/google/services/sql/resource_sql_database_instance.go#L1811-L1843)), so `terraform apply` on a MySQL `google_sql_database_instance` would fail. Would you be open to having the plane log in through `root@localhost` (over the socket) or a dedicated admin account of its own, so `root@%` can be an ordinary account that can be deleted and recreated? That would flip `mysqlSystemDatabasesAndRootCannotBeDeleted` and the `root`/`%` assert near line 426.

I checked the host handling (`%` default on MySQL, rejected on PostgreSQL) and the version set against the [SQL Admin v1 discovery document](https://www.googleapis.com/discovery/v1/apis/sqladmin/v1/rest); those look right. Once `root@%` can be deleted, there are no other blockers from my side.

avison9:
Right, and thanks for the pointer into the provider. Addressed in 3d18cff:

- **`root@%` is an ordinary account now.** The plane logs in over the container's Unix socket (`mysql --protocol=socket --socket=/var/run/mysqld/mysqld.sock`), which the server authenticates as `root@localhost`; it used to connect over TCP to 127.0.0.1, which with the image's `skip-name-resolve` authenticates as `root@%`, and that is why `root@%` had to be protected. `root@localhost` is the only reserved identity left (`CloudSqlEngine.isReservedIdentity`). `users.delete` on `root@%` runs `DROP USER`, `users.insert` recreates it with the requested password, and `users.update` rotates it on the server instead of being acknowledged silently. `127.0.0.1` and `::1` are no longer reserved either: the image only provisions `root@localhost` and `root@%`, and the socket login never resolves to them.
- **Checked against a real `mysql:8.4.11` container**, the same sequence the provider runs: `SELECT CURRENT_USER()` is `root@localhost` over the socket and `root@%` over TCP; after `DROP USER 'root'@'%'` the socket login keeps working, including across a container restart, and `CREATE USER 'root'@'%' ... ; ALTER USER ...` plus the per-database `GRANT` recreate it. The TCP readiness probe (`mysqladmin ping ... -u root`) still exits 0 once `root@%` is gone (the server answers "Access denied", which `mysqladmin ping` documents as alive), so a persisted instance restarts normally with that account deleted.
- **Tests flipped as you described:** `mysqlSystemDatabasesAndRootCannotBeDeleted` is now `mysqlSystemDatabasesCannotBeDeletedButRootAtPercentCan` (delete, `NOT_FOUND`, reinsert); `onlyTheProvisionedRootIdentityIsProtected` asserts the plane receives the drop, the recreate and a password rotation for `root@%`; `localRootIdentitiesAreReservedOnMysql` is `rootAtLocalhostIsReservedOnMysql`. The plane test asserts every `mysql` invocation carries `--protocol=socket` and never `127.0.0.1`, and lists the exact SQL issued for `root@%`. The REST test runs the provider's delete then insert; `CloudSqlAdminTest` in `sdk-test-java` does the same through the SDK (`users().delete(...).setName("root").setHost("%")`, then `insert`).
- Docs bullet in `cloud-sql-postgres.md` updated to match.

Rebased onto `main` (the docs refresh in #190 moved the README and compat tables; resolved by folding the MySQL rows into the new layout). Full suite 1117/0/0/0.

hectorvent:
Thank you, that is a better fix than the one I suggested: logging in over the socket as `root@localhost` removes the reason `root@%` had to be protected at all, rather than working around it.

I re-checked the provider source: it lists users right after create and deletes the one whose host is `%`, so an ordinary, deletable `root@%` is exactly what it needs. `users.insert` recreating it and `users.update` rotating the password on the server both line up.

I also ran a range diff against the previous head: the other four commits are unchanged apart from the rebase onto main, where the README and compatibility README hunks landed cleanly on the refreshed docs. Cloud SQL tests pass locally, 39.

No blockers from my side.

hectorvent (approved):
Approving: root@% blocker fixed at the source (socket login), provider behaviour re-verified, CloudSql* 39/0 locally.

PR #204, opened 2026-09-14
feat(cloudsql): list the built-in postgres user on PostgreSQL instances

### LINKED ISSUE #203: [FEAT] Cloud SQL: list the built-in postgres user on PostgreSQL instances
## Service

Cloud SQL

## API Method / Feature

`sqladmin` v1 `users.list` / `users.get` on a PostgreSQL instance: the built-in `postgres` account the instance is provisioned with.

## GCP Documentation

- https://cloud.google.com/sql/docs/postgres/users ("When you create a new Cloud SQL for PostgreSQL instance, the default admin user `postgres` is created")
- `users.list`: https://cloud.google.com/sql/docs/postgres/admin-api/rest/v1/users/list
- `User.type` (`BUILT_IN`): https://cloud.google.com/sql/docs/postgres/admin-api/rest/v1/users#SqlUserType

## Why is this needed?

A fresh PostgreSQL instance on floci-gcp answers `users.list` with an empty list, while real Cloud SQL lists the `postgres` user (`type: BUILT_IN`). Code that inspects or imports the default user (Terraform `google_sql_user` import, scripts that rotate the `postgres` password through `users.update`) sees a different world locally: `users.get` for `postgres` is `404`, and `users.update` cannot be issued against it.

#202 introduced the engine model and lists `root@%` on MySQL instances this way; the PostgreSQL side was left as it was to keep that PR to MySQL. This is the agreed follow-up.

## Proposed scope

- Seed a `postgres` user (`type: BUILT_IN`) on `instances.insert` for `POSTGRES_*` instances; `users.get` returns it, `users.delete` refuses it with `400 FAILED_PRECONDITION` (it is the emulator's admin login), matching the MySQL treatment of `root`.
- `users.update` on it is acknowledged without changing the server password, as for MySQL `root` (the plane's own login stays `postgres`/`postgres`; documented).
- Existing instances persisted without the user gain it on restart so `users.list` is consistent after an upgrade.
- Tests at service and REST level; compat `CloudSqlAdminTest` asserts `postgres` is listed.

Depends on #202.

- [x] Yes
- [ ] No

avison9:
> **Stacked on #202. Merge #202 first.** This branch is #202 plus one commit; until #202 lands, the diff here shows both. Once #202 merges, GitHub will reduce this PR to its own commit (`feat(cloudsql): list the built-in postgres user on PostgreSQL instances`) and it merges cleanly. Opened now rather than after #202 so both can be reviewed in one sitting.

## Summary

A fresh PostgreSQL instance answered `users.list` with an empty list, while real Cloud SQL lists the `postgres` role the instance is provisioned with (`type: BUILT_IN`). #202 gave MySQL instances their `root@%` this way and deliberately left PostgreSQL alone; this is that follow-up. Closes #203. Stacked on #202 (uses `CloudSqlEngine.builtInUser`).

```bash
# BEFORE
H=http://localhost:4588/v1/projects/p
curl -s -X POST "$H/instances" -H 'Content-Type: application/json' -d '{"name":"pg-main","databaseVersion":"POSTGRES_18"}' >/dev/null
curl -s "$H/instances/pg-main/users" | jq '.items'
# []
curl -s "$H/instances/pg-main/users/postgres" | jq .error.status
# "NOT_FOUND"

# AFTER
curl -s "$H/instances/pg-main/users" | jq '[.items[] | {name, type}]'
# [{"name":"postgres","type":"BUILT_IN"}]
curl -s -X DELETE "$H/instances/pg-main/users?name=postgres" | jq .error.message
# "Built-in user cannot be deleted: postgres"
```

### What changed

- `CloudSqlEngine.POSTGRES.builtInUser` is now `postgres`; the seeding, `users.get`, and the delete refusal already exist from #202 and apply unchanged.
- `CloudSqlPostgresDataPlane` leaves the `postgres` role alone on `createOrUpdateUser` and `grantDatabaseAccess`, the way the MySQL plane treats `root`: it is the plane's own `psql` login, so rotating it would strand every later DDL call. `users.update` on it is still accepted at the API level.
- `CloudSqlService.backfillBuiltInUsers()` runs at startup (data plane or mock) so instances persisted by an earlier build list the user after an upgrade. Idempotent; never overwrites an existing record. It writes into each instance's own project, since startup has no request context.
- Docs: one paragraph in `cloud-sql-postgres.md`.

Verified on a real `postgres:18.4-alpine` container: `users.list` shows `postgres`/`BUILT_IN`; after `users.update` on it, `databases.insert` and `users.insert` still work and the new user connects over `psql`; delete refused; instance delete clean.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

`users.list` on a Cloud SQL for PostgreSQL instance returns the `postgres` user with `type: BUILT_IN` (SQL Admin v1 `SqlUserType`; "When you create a new Cloud SQL for PostgreSQL instance, the default admin user postgres is created"). Compat `CloudSqlAdminTest` now asserts it on every PostgreSQL version it creates.

## Tests

- `CloudSqlServiceTest` (+2): the user is listed as `BUILT_IN` with no host, `users.get` finds it, `users.update` is accepted and stays password-free, `users.delete` is `400 FAILED_PRECONDITION`; startup backfill adds it to a legacy instance record, is idempotent, and preserves an existing record. Both fail on the parent branch (`expected: <1> but was: <0>`).
- Existing `dataPlaneReceivesInstanceDatabaseAndUserLifecycleEvents` gains the `grant:appdb:postgres` event the service now emits (the plane itself skips it).
- `CloudSqlRestIntegrationTest`: asserts the `BUILT_IN` entry and that no listed user carries a password.
- Compat `CloudSqlAdminTest`: PostgreSQL branch asserts `postgres` is listed as `BUILT_IN`.

Full suite: 1077 run / 0 failures / 0 errors / 0 skipped on top of the #202 branch (1075), i.e. the 2 new tests and nothing else.

## Checklist


avison9:
Fair point, and it belongs to the #202 code rather than this commit, so it is fixed there: b795193 validates the full `databaseVersion` in `CloudSqlEngine` (POSTGRES_15 to 18; MYSQL_8_0, MYSQL_8_0_NN, MYSQL_8_4), so mock mode and the Docker path accept exactly the same set. This branch is restacked on top; its own commit is unchanged.

hectorvent:
Thanks for splitting this out from #202 and for stating the stacking up front, it made the own commit easy to read.

1. **(blocking)** Since this branch carries #202's code, the `root@%` point on #202 applies here as well; a rebase once it is settled should be enough.
2. **(blocking)** `backfillBuiltInUsers()` re-creates the built-in user whenever the record is absent, so once a built-in user can be deleted (the Terraform provider deletes `root@%` after every MySQL create) each restart would bring it back. Would a one-time marker on the instance record, instead of a presence check, work for you?

Listing `postgres` as `BUILT_IN` and accepting `users.update` on it matches how the Terraform provider rotates the PostgreSQL root password by name, so that part looks right to me. No other blockers from my side.

avison9:
Both addressed, thanks.

1. **`root@%`**: rebased onto #202 at 3d18cff, where the plane now logs in over the Unix socket as `root@localhost` and `root@%` is an ordinary, deletable account. Nothing else needed here for that.

2. **Backfill marker** in 6a4e528: `createBuiltInUser` is now `seedBuiltInUser`, once per instance. It records a marker and skips instances that carry one, so a `root@%` the API deleted stays deleted across restarts, while an instance persisted before built-in users existed is still seeded on its first startup after the upgrade (and marked then). Two small choices worth stating:
   - The marker is a sibling key in the instance store (`built-in-user-seeded/<instance>`) rather than a field on the instance record, so it never appears in `instances.get` or `instances.list`. `deleteInstance` removes it with the instance, so recreating an instance under the same name is seeded again.
   - It is written for new instances too, not only by the backfill, so the presence check is gone entirely.

   `backfillSeedsOnceAndNeverResurrectsABuiltInUserTheApiDeleted` covers the legacy instance (seeded on first backfill, deleted, absent after a simulated restart), a freshly created one, the record and listing staying clean, and re-seeding after delete and recreate. The existing idempotency test still passes.

Full suite 1120/0/0/0. Merge order unchanged: #202 first.

hectorvent:
Thanks, both of these are fixed exactly as described.

Keeping the marker as a sibling key rather than a field on the instance record is the right call: `instances.get` keeps its shape, and I checked that every enumeration filters on the `instances/` prefix, so the marker cannot show up as a phantom instance in `instances.list`. `deleteInstance` clearing it means a recreated instance seeds again, which is what you want.

Seeding instances persisted before built-in users existed, once, on their first startup after the upgrade, is also the behaviour I would have picked. Cloud SQL tests pass locally, 39.

One mechanical thing before this can land: I merged #202 just now, and since we squash merge, its four commits came in as one. This branch still carries the originals, so it currently conflicts in five files (`CloudSqlEngine.java` as an add/add, plus `CloudSqlService`, `CloudSqlPostgresDataPlane` and two tests). A rebase onto main should drop them and leave just your two commits. Sorry for the ordering, that is our merge style rather than anything you did.

No review blockers from my side; it is just the rebase.

hectorvent:
Rebase looks right: a range diff against the previous head shows #202's five commits dropped, now that they are on main as one squashed commit, and your own two unchanged. Cloud SQL tests pass locally, 39.

Thanks for turning that around, and sorry again for the ordering.

No blockers from my side.

hectorvent (approved):
Approving: rebase verified by range-diff, CloudSql* 39/0 locally.

PR #214, opened 2026-09-16
fix(gcs): publish composed objects atomically

snazy:
## Summary

- finalize composite metadata before its single storage publication
- keep destination publication under the bucket and object locks
- add a deterministic regression test for concurrent bucket deletion

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

Composite objects continue to report their accumulated component count without an MD5 hash. The fix removes the second metadata write that could race with object and bucket deletion and ensures finalize notifications observe the final metadata.

Verified with GcsServiceTest, GcsGrpcControllerTest, and GcsComposeLimitsRestIntegrationTest.

## Checklist

- [ ] `./mvnw test` passes locally (focused GCS tests pass)
- [ ] New or updated integration test added (deterministic service-level concurrency regression added)

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Checklist

- [ ] `./mvnw test` passes locally
- [ ] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)


hectorvent:
Thank you, this is a tidy fix, and the second commit turning the timeout into an observed `BLOCKED` state makes the regression test genuinely deterministic (it fails against main's `GcsService`, passes here).

Writing the composite metadata once, under the bucket and object locks, also means the finalize notification now carries the final shape (componentCount set, no md5Hash), which matches the [storage v2 proto](https://github.com/googleapis/googleapis/blob/aa87617d67/google/storage/v2/storage.proto#L2813-L2819): "composite objects provide only crc32c hashes". Both the REST and gRPC compose paths go through this method, so one change covers both.

No blockers from my side.

hectorvent (approved):
Approving: re-verified on the merge result against current main, GCS suite 264/0.

PR #219, opened 2026-09-16
feat(kafka): implement the ManagedKafka ACL RPCs

### LINKED ISSUE #215: [FEAT] Managed Kafka: implement the ACL RPCs (acls/** resource, addAclEntry, removeAclEntry)
## Service

Managed Kafka

## API Method / Feature

`google.cloud.managedkafka.v1.ManagedKafka` ACL RPCs: `ListAcls`, `GetAcl`, `CreateAcl`, `UpdateAcl`, `DeleteAcl`, `AddAclEntry`, `RemoveAclEntry`, reached over REST as:

```
GET    /v1/{parent=projects/*/locations/*/clusters/*}/acls
GET    /v1/{name=projects/*/locations/*/clusters/*/acls/**}
POST   /v1/{parent=projects/*/locations/*/clusters/*}/acls?aclId=...
PATCH  /v1/{acl.name=projects/*/locations/*/clusters/*/acls/**}
DELETE /v1/{name=projects/*/locations/*/clusters/*/acls/**}
POST   /v1/{acl=projects/*/locations/*/clusters/*/acls/**}:addAclEntry
POST   /v1/{acl=projects/*/locations/*/clusters/*/acls/**}:removeAclEntry
```

## GCP Documentation

- `managed_kafka.proto` (service and request messages): https://github.com/googleapis/googleapis/blob/master/google/cloud/managedkafka/v1/managed_kafka.proto
- `resources.proto` (`Acl`, `AclEntry`, the `acl_id` grammar): https://github.com/googleapis/googleapis/blob/master/google/cloud/managedkafka/v1/resources.proto#L531
- REST reference: https://cloud.google.com/managed-service-for-apache-kafka/docs/reference/rest/v1/projects.locations.clusters.acls
- Concepts: https://cloud.google.com/managed-service-for-apache-kafka/docs/authentication-authorization

## Why is this needed?

Clusters, topics and consumer groups are served; ACLs are the one remaining resource in the core `ManagedKafka` service, so today every call to the seven RPCs above is a 404. Terraform `google_managed_kafka_acl`, `gcloud managed-kafka acls`, and any client that provisions per-topic access next to the topic itself has no local target, and the docs list of supported operations cannot claim the full v1 surface until this lands.

## Proposed scope

- `Acl` resource stored per cluster, keyed by the multi-segment `acl_id`, which encodes the Kafka resource pattern and must be parsed and validated: `cluster`; `topic/{name}`, `consumerGroup/{name}`, `transactionalId/{name}`; `topicPrefixed/{name}`, `consumerGroupPrefixed/{name}`, `transactionalIdPrefixed/{name}`; `allTopics`, `allConsumerGroups`, `allTransactionalIds`. Anything else is `400 INVALID_ARGUMENT`. The derived output-only fields `resource_type`, `resource_name`, `pattern_type` are populated from the id.
- `acl_entries` (principal, permission_type, operation, host) capped at 100 per ACL as the proto states; `AddAclEntry` / `RemoveAclEntry` mutate the set and return the updated `Acl`, with `RemoveAclEntry` reporting `acl_deleted: true` when the last entry goes.
- `etag` generated on create and get, required and checked on `UpdateAcl` (`412`/`ABORTED` on mismatch per the proto's concurrency note).
- Standard list pagination shape (`acls`, `nextPageToken`).
- Data plane: decide at implementation whether entries are applied to the Redpanda container via `rpk security acl` (Redpanda supports Kafka ACLs) or kept metadata-only like other authorization surfaces in the emulator. Metadata-only is the minimum; applying them is the parity option and should be behind the existing mock flag.
- Docs: add the seven RPCs to `docs/services/managed-kafka.md` (which is also missing `UpdateConsumerGroup`, already implemented) and compat coverage in `sdk-test-java` `KafkaTest`.

- [x] Yes
- [ ] No

avison9:
## Summary

Implements the ACL half of `google.cloud.managedkafka.v1.ManagedKafka`: `ListAcls`, `GetAcl`, `CreateAcl`, `UpdateAcl`, `DeleteAcl`, `AddAclEntry`, `RemoveAclEntry`. Clusters, topics and consumer groups were already served; with this the emulator covers the service's full v1 RPC list. Closes #215.

```bash
# BEFORE
B=http://localhost:4588/v1/projects/p/locations/us-central1/clusters/c
curl -s -o /dev/null -w '%{http_code}\n' -X POST "$B/acls?aclId=topic/orders" -H 'Content-Type: application/json' \
  -d '{"aclEntries":[{"principal":"User:svc@p.iam.gserviceaccount.com","permissionType":"ALLOW","operation":"READ","host":"*"}]}'
# 404

# AFTER
curl -s -X POST "$B/acls?aclId=topic/orders" -H 'Content-Type: application/json' \
  -d '{"aclEntries":[{"principal":"User:svc@p.iam.gserviceaccount.com","permissionType":"allow","operation":"read","host":"*"}]}'
# {"name":".../clusters/c/acls/topic/orders","aclEntries":[{"principal":"User:svc@...","permissionType":"ALLOW","operation":"READ","host":"*"}],
#  "etag":"5f250017e05c401e","resourceType":"TOPIC","resourceName":"orders","patternType":"LITERAL"}
curl -s -X POST "$B/acls/allTopics:addAclEntry" -H 'Content-Type: application/json' \
  -d '{"principal":"User:*","permissionType":"ALLOW","operation":"DESCRIBE","host":"*"}'
# {"acl":{...,"resourceName":"*"},"aclCreated":true}
curl -s -X POST "$B/acls/allTopics:removeAclEntry" -H 'Content-Type: application/json' -d '<same entry>'
# {"aclDeleted":true}
```

### What changed

- **`AclResourcePattern`** (new): parses the `acl_id` grammar from `resources.proto` (`cluster`; `topic|consumerGroup|transactionalId/{name}`; the `*Prefixed/{name}` forms; `allTopics|allConsumerGroups|allTransactionalIds`) into the output-only `resourceType` / `resourceName` / `patternType`, and rejects anything else with `400`. Case-sensitive, as the API is.
- **`StoredAcl`, `AclEntry`** (new models): entry identity is all four fields, which is what `addAclEntry` deduplicates on and `removeAclEntry` matches on.
- **`KafkaService`**: the seven operations over a new `kafka-acls.json` store; cluster delete cascades ACLs. Entry rules from the proto: `User:` prefix on the principal, `ALLOW`/`DENY`, the eleven Kafka operations, `host` must be `*`; permission and operation accepted case-insensitively and stored upper-case; 100 entries max. `etag` rotates on every write; `UpdateAcl` with a stale one is `409 ABORTED` (AIP-154), without one it is unconditional; `updateMask` may only name `aclEntries`. `AddAclEntry` creates a missing ACL (`aclCreated`), `RemoveAclEntry` deletes an emptied one (`aclDeleted`), matching the response messages' shapes; the remove response honours the `oneof`.
- **`KafkaController`**: seven routes. The id is bound as `{aclId: [^:]+}` so a multi-segment id is captured whole while `.../acls/topic/orders:addAclEntry` still reaches the custom method. `ListAcls` paginates with `pageSize`/`pageToken` as the proto declares.
- Docs: ACL section in `managed-kafka.md`, and `UpdateConsumerGroup` added to the supported list (it was implemented but undocumented).

### Design notes

- ACLs are control-plane metadata. Topics already are in this emulator (no `rpk` call), and the Redpanda container runs without an authorizer, so applying entries would gate nothing. Documented as such. Wiring them into Redpanda's `rpk security acl` is possible later without changing the API surface.
- Removing an entry that is not present is `404 NOT_FOUND`; adding one that is already present is a no-op returning the unchanged ACL. The proto does not pin either case; both are the conservative reading.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

Wire shapes from `googleapis/googleapis` master: `managed_kafka.proto` (the seven RPCs, their `google.api.http` bindings including `acls/**` and the `:addAclEntry` / `:removeAclEntry` custom methods, `AddAclEntryResponse.acl_created`, `RemoveAclEntryResponse` oneof) and `resources.proto` (`Acl`, `AclEntry`, the `acl_id` grammar and field rules). Verified with the compat suite's `KafkaTest` against a running instance (the suite drives Managed Kafka over raw REST, as the existing cases do): 12/12.

## Tests

- `AclResourcePatternTest` (new, 3): every documented spelling derives its pattern; fifteen malformed ids are `400`; names keep their characters.
- `KafkaAclRestIntegrationTest` (new, 3, `@QuarkusTest`): the full lifecycle over REST (create with canonicalised case, get and list through the multi-segment id, update with etag rotation and stale-etag `409`, `addAclEntry` on existing and missing, `removeAclEntry` returning the ACL or `aclDeleted`, delete, cluster cascade); id and entry validation; the 100-entry cap on both create and add. Written first: 3/3 failed with 404 before the change.
- Compat `KafkaTest` +1 (`aclsFollowTheResourcePatternGrammarAndSupportIncrementalEntries`), README counts updated.

Full suite: baseline on `main` @ `2ecf7d2` (0.9.0) is 1058 run / 0 failures / 0 errors / 0 skipped; with this change 1064 / 0 / 0 / 0 (the 6 new tests, nothing else changed).

## Checklist


hectorvent:
Thank you for this, and for waiting while the duplicate was sorted out.

I checked the whole surface against the proto rather than spot checking: the seven RPCs and their http bindings match, the `acl_id` grammar covers exactly the documented forms including the three wildcards, `GROUP` is the right spelling for the consumer group patterns, and the 100 entry cap is enforced before the entry is added rather than after, so a rejected add leaves the stored acl untouched.

The two entry RPCs return their own messages, and `RemoveAclEntry` honouring the oneof (either `acl` or `aclDeleted`, never both) with the acl deleted when its last entry goes is exactly what the proto describes.

One question rather than a finding: removing an entry that is not on the acl returns NOT_FOUND. The proto does not say what real Managed Kafka does there, so if you happen to know, it would be worth a line in the docs page either way.

No blockers from my side.

hectorvent (approved):
Approving: seven RPCs, acl_id grammar, GROUP spelling, 100-entry cap and both entry response messages verified against the proto.

avison9:
Greptile's finding is right, and it matches the "Concurrency and Storage Invariants" section that landed in CONTRIBUTING.md this morning, so this is addressed in 4f15a20 the way `SecretManagerService` does it:

- **Invariant:** an ACL exists only under a cluster that exists, and no ACL mutation loses another's entry.
- **Competing paths:** `DeleteCluster` (ACL cascade) against `CreateAcl`, `UpdateAcl`, `DeleteAcl`, `AddAclEntry`, `RemoveAclEntry`, `GetAcl`, `ListAcls`; and any two of the mutations against each other on one ACL (`AddAclEntry` / `RemoveAclEntry` / `UpdateAcl` are read-modify-write).
- **Serialization:** `KafkaService` now holds one lock per cluster (striped on the cluster name, 256 stripes). `deleteCluster` and all seven ACL operations run inside it, and the cascade deletes ACLs before the cluster record, so no ACL operation can see the cluster and then publish under a parent that is gone.
- **Lock order:** the service takes exactly one lock per operation, so no path acquires two and there is no order to preserve.
- **Deferred, stated:** topic and consumer group paths predate this PR and are not serialized with deletion (`createTopic` can still race `deleteCluster`). They are untouched here; the lock is in place for a follow-up to wrap them.
- **Tests:** `KafkaAclConcurrencyTest`, deterministic (latches inside the ACL store, no sleeps), two cases:
  1. `clusterDeletionSerializesEveryAclOperationAndLeavesNoOrphan`: the delete blocks inside the first ACL `delete` while holding the lock; eight competing ACL operations (including the two that create an ACL when it is absent) must not complete; after release all eight observe `NOT_FOUND` and the ACL store is empty.
  2. `concurrentEntryMutationsOnOneAclNeverLoseAWrite`: the first `addAclEntry` blocks inside `put` after reading and appending; a second `addAclEntry` and a `removeAclEntry` must not have started their read; after release the ACL holds exactly `User:b` and `User:c`.
  Both fail when the lock is replaced by a fresh object per call (checked), so they detect the races they describe.

A package-private constructor over the four stores was added so the service can be built in a unit test; the injected constructor delegates to it. Full suite 1107/0/0/0.

PR #222, opened 2026-09-19
fix(docker): update native Dockerfile runtime base to ubi9-minimal

### LINKED ISSUE #221: [BUG] Dockerfile.native build fails due to missing /bin/sh shell in ubi9-micro base image
## Service

Docker / Build Infrastructure

## GCP API / Method

N/A - Docker container build

## Expected behavior

`docker build -f docker/Dockerfile.native .` should build the native runtime image without errors.

## Actual behavior

Stage 2 of `docker/Dockerfile.native` uses `registry.access.redhat.com/ubi9-micro:9.7`. Because `ubi9-micro` lacks `/bin/sh` and system utilities, the user setup step `RUN echo 'floci:x:1001:0:Floci:/app:/sbin/nologin' >> /etc/passwd ...` fails during `docker build` with:

```text
exec /bin/sh: no such file or directory
```

## Reproduction

```bash
docker build -t floci-gcp:native -f docker/Dockerfile.native .
```

## Environment

- floci-gcp version / image tag: main
- Base image: `registry.access.redhat.com/ubi9-micro:9.7`

kanishkkumarsingh2004:
## Summary

Fixes native container image builds where `docker/Dockerfile.native` failed in Stage 2 due to `registry.access.redhat.com/ubi9-micro:9.7` lacking `/bin/sh`. Switched the Stage 2 runtime base image to `registry.access.redhat.com/ubi9-minimal:9.7`. Also updated `docker-compose.yml` build context.

Closes #221

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

Fixes container build infrastructure for native images. Verified `ubi9-minimal:9.7` layer retrieval and build steps.

## Checklist

- [x] Tested build step locally


hectorvent:
Thanks for chasing this one down, and for linking the issue with a clean repro command.

**(blocking)** The `docker-compose.yml` hunk swaps `docker/Dockerfile.native` for `docker/Dockerfile`, which is the JVM build. That compose file is what the local compat stack runs against (`Makefile:75`), so local runs would quietly stop exercising the native binary. Greptile flagged the same thing above. Would you drop that hunk and keep the change to `docker/Dockerfile.native` only?

**(blocking)** I could not reproduce the missing shell, so I have asked for a few environment details on #221. On `ubi9-micro:9.7` I get `/bin/sh -> bash` on amd64 and arm64, the stage 2 `RUN` block runs, and stage 2 builds on both platforms. The same base and line build in every release through `docker/Dockerfile.native-package`.

**(worth knowing)** `ubi9-minimal:9.7` is 38.2 MB against ubi9-micro's 6.7 MB, so this swap adds roughly 31 MB to an image we recently shrank. If your build failure turns out to be real, a fix that keeps ubi9-micro would be preferable.

Once we see the build output we will know which way to go.

hectorvent:
@kanishkkumarsingh2004 Closing issue, 
Lets track any comment in the issue directly

PR #225, opened 2026-09-19
fix(cloudrun): stream invocation responses

### LINKED ISSUE #224: [BUG] Cloud Run invocation proxy buffers streaming responses and breaks SSE
## Service

Cloud Run invocation proxy

## Expected behavior

Streaming HTTP responses from an emulated Cloud Run service should be forwarded incrementally to the caller. In particular, a Server-Sent Events response should deliver its headers and each event as soon as the upstream service flushes them, while keeping the connection open.

This should support standard streaming response types such as `text/event-stream`, NDJSON, and other chunked responses.

## Actual behavior

Floci waits for the upstream response body to complete before sending any response headers or body bytes to the caller. An SSE response normally never completes, so clients remain in the connecting state and receive no events.

The behavior was reproduced with an upstream service that:

1. Returns `Content-Type: text/event-stream`.
2. Flushes the response headers.
3. Immediately writes `: connected\n\n`.
4. Keeps the connection open and sends periodic heartbeats.

A direct request to the Cloud Run runtime container receives `: connected` immediately:

```console
$ timeout 3 curl --no-buffer http://<runtime-container-ip>:<port>/events
: connected
```

The equivalent request through the Floci Cloud Run service URL receives neither response headers nor body data before timing out:

```console
$ timeout 20 curl --no-buffer --verbose http://<service>.us-central1.run.localhost.floci.io:4588/events
> GET /events HTTP/1.1
> Host: <service>.us-central1.run.localhost.floci.io:4588
>
# No response headers or body are received.
```

Browser `EventSource` exhibits the same behavior: `onopen` is never called.

## Root cause

In `CloudRunInvocationController.proxy()`, the upstream request is executed synchronously with `HttpResponse.BodyHandlers.ofByteArray()`:

https://github.com/floci-io/floci-gcp/blob/2ecf7d25e5df5b47cca1149e66a750b8454b15bf/src/main/java/io/floci/gcp/services/cloudrun/CloudRunInvocationController.java#L184-L193

```java
HttpResponse<byte[]> response = httpClient.send(
        request,
        HttpResponse.BodyHandlers.ofByteArray());
return toResponse(response);
```

`ofByteArray()` buffers the complete upstream body before `send()` returns. `toResponse()` therefore cannot send the status, headers, or body until the upstream response ends. This is incompatible with indefinite response streams and can also cause large finite responses to be fully buffered in memory.

## Suggested fix

Proxy the upstream body as a stream rather than materializing it as a byte array. One possible implementation is `HttpResponse.BodyHandlers.ofInputStream()` combined with a JAX-RS `StreamingOutput` response entity that copies and flushes bytes incrementally.

The implementation should also:

- Close or cancel the upstream response when the downstream client disconnects.
- Preserve the existing hop-by-hop header filtering.
- Avoid adding a `Content-Length` header to streamed responses.
- Keep request timeout behavior compatible with long-lived Cloud Run streaming responses.

## Suggested regression test

Add a Cloud Run invocation integration test whose upstream handler writes and flushes one chunk, waits on a latch, and only then completes. Assert that the caller receives the response status, headers, and first chunk before releasing the latch. An SSE-specific test can additionally verify that the connection remains open while multiple events are delivered.

## Environment

- Floci image: `floci/floci-gcp:0.9.0`
- Revision: `2ecf7d25e5df5b47cca1149e66a750b8454b15bf`
- Runtime: Docker on Linux

mheras:
## Summary

- stream Cloud Run invocation response bodies instead of buffering them
- flush each copied chunk to support SSE and other long-lived responses
- close the upstream response stream when copying completes or the downstream disconnects
- add a regression test proving the first chunk arrives before the upstream response completes

## Verification

- full Maven test suite: 1055 tests, 0 failures
- end-to-end SSE through the Floci Cloud Run service URL returns `200`, `text/event-stream`, and `: connected` while the response remains open

Closes #224


hectorvent:
Thank you, this is a clean fix and the regression test that asserts the first chunk arrives before the upstream finishes is exactly the right thing to pin.

**(blocking)** CI has run now and everything passes except the commit message gate: `357e7d2` is git's default revert subject, `Revert "fix(cloudrun): proxy runtimes over HTTP/1.1"`, and the workflow wants a conventional type, so `revert: proxy runtimes over HTTP/1.1` would pass. Since the commit it reverts is also on this branch, the simplest route is to drop both with an interactive rebase and force push, leaving just the streaming commit.

**(follow-up, separate PR)** With `BodyHandlers.ofInputStream()` the request timeout now covers only the response headers, so a runtime that sends headers and then stalls holds the invocation open indefinitely, where the old byte array handler bounded the whole body. Greptile raises the same point. Streaming has to outlive a request timeout, so I do not think this blocks the fix, but a cap on total duration for non streaming responses would be worth a follow-up.

The header handling is fine: `content-length` and `transfer-encoding` are already dropped as hop by hop, so chunked responses stay consistent. Ran the Cloud Run tests locally too, 54 passing.

PR #227, opened 2026-09-20
fix(cloudrun): match runtime protocol contract

### LINKED ISSUE #226: [BUG] Cloud Run runtime proxy attempts h2c by default
## Service

Cloud Run invocation proxy

## Expected behavior

Floci should match the Cloud Run container protocol contract:

- Send HTTP/1.1 to the ingress container by default.
- Use cleartext HTTP/2 (`h2c`) only when the ingress container port is explicitly named `h2c`.

## Actual behavior

`CloudRunInvocationController` constructs `java.net.http.HttpClient` without specifying a version. The client prefers HTTP/2 and attempts an h2c upgrade for cleartext runtime endpoints even when the Cloud Run service did not request end-to-end HTTP/2.

This breaks otherwise valid HTTP/1.1 ingress servers. It was reproduced with a Next.js standalone server listening directly on `0.0.0.0:8080`. Adding an Nginx protocol adapter made the service work, but that introduces a local-only architecture difference that is unnecessary on real Cloud Run.

## Root cause

The service and revision protobufs preserve `containers[0].ports[0].name`, but `CloudRunRuntimeService` only copies `containerPort` into `CloudRunRuntimeInstance`. The invocation controller therefore cannot distinguish the default protocol from an explicitly configured h2c service.

## Suggested fix

- Preserve whether the ingress port name is `h2c` in `CloudRunRuntimeInstance`.
- Set each proxied `HttpRequest` to HTTP/1.1 by default.
- Set it to HTTP/2 only for runtimes whose ingress port is explicitly named `h2c`.
- Add regression coverage for both modes.

## References

Google Cloud Run documentation states that HTTP/2 requests are downgraded to HTTP/1 before reaching the container by default. End-to-end HTTP/2 must be enabled explicitly by naming the container port `h2c`.

## Environment

- Floci base version: `0.9.0`
- Runtime: Docker on Linux

mheras:
## Summary

- use HTTP/1.1 for Cloud Run ingress containers by default
- preserve explicit `ports.name = "h2c"` on runtime instances
- use HTTP/2 only for runtimes explicitly configured for h2c
- preserve HTTP/1.1 behavior when loading runtime records created before this field existed
- add regression coverage for protocol detection and request selection

## Verification

- focused Cloud Run tests: 17 tests, 0 failures
- full Maven test suite: 1056 tests, 0 failures
- validated a Next.js standalone ingress container directly on `0.0.0.0:8080`, without a protocol-adapter reverse proxy

Closes #226


hectorvent:
Thanks, this one matches the spec closely. I checked `ContainerPort.name` against the proto, which says ["Allowed values are "http1" and "h2c""](https://github.com/googleapis/googleapis/blob/aa87617d67/google/cloud/run/v2/k8s.min.proto#L185), so defaulting to HTTP/1.1 and switching only on an explicit `h2c` port name is right, and it fixes the previous behaviour where the JDK client negotiated HTTP/2 for every runtime.

**(follow-up, separate PR)** As Greptile notes, the new test asserts the request version rather than a real h2c exchange, so protocol selection is pinned but the wire itself is not. That is awkward to test without an h2c upstream, so it seems like a reasonable thing to leave for later if you agree.

Keeping the older stored records on HTTP/1.1 through the extra constructor is a nice touch. Ran the Cloud Run tests locally, 55 passing. No blockers from my side.

hectorvent (approved):
Approving: ContainerPort.name checked against the proto, CloudRun* 55/0 locally, 12/12 green.

Issue #228, opened 2026-09-20
[BUG] GKE returns 11 operationType values that container.v1 Operation.Type does not declare


hectorvent:
## Service

GKE (Kubernetes Engine)

## GCP API / Method

`container.v1.ClusterManager`, the `Operation.operationType` field returned by most mutating RPCs.

## Expected behavior

`Operation.Type` is a closed enum. [cluster_service.proto:3984-4107](https://github.com/googleapis/googleapis/blob/aa87617d67/google/container/v1/cluster_service.proto#L3984) declares exactly 19 values, from `TYPE_UNSPECIFIED = 0` to `FLEET_FEATURE_UPGRADE = 19`. Every Operation an SDK receives must carry one of them, because a generated client parses the field into that enum.

## Actual behavior

`OperationType` declares 23 values, and 11 of them are not in the proto:

```
UPDATE_NODE_POOL            SET_ADDONS_CONFIG        SET_LOGGING_SERVICE
SET_MONITORING_SERVICE      SET_LOCATIONS            SET_NODE_POOL_AUTOSCALING
SET_LEGACY_ABAC             START_IP_ROTATION        COMPLETE_IP_ROTATION
COMPLETE_NODE_POOL_UPGRADE  ROLLBACK_NODE_POOL_UPGRADE
```

They are emitted verbatim (`src/main/java/io/floci/gcp/services/gke/operations/OperationType.java`, read back through `StoredOperation.getOperationType()` and serialized by `KubernetesController`), so `UpdateNodePool`, `SetNodePoolAutoscaling`, `SetNodePoolSize`, `SetLocations`, `StartIPRotation` and the rest hand back an Operation outside the wire contract.

Nothing catches it today: `GkeServiceTest` asserts against floci's own enum, and the compatibility suite only ever checks `UPGRADE_MASTER`.

## Reproduction

```bash
# 11 values floci can emit that the proto does not declare
grep -oE "^\s+[A-Z_]+" src/main/java/io/floci/gcp/services/gke/operations/OperationType.java \
  | tr -d ' ' | grep -E "^[A-Z_]+$" | sort > /tmp/floci.txt
sed -n '3984,4107p' local/google/googleapis/google/container/v1/cluster_service.proto \
  | grep -oE "^\s+[A-Z_]+ = [0-9]+" | awk '{print $1}' | sort > /tmp/proto.txt
comm -13 /tmp/proto.txt /tmp/floci.txt
```

## Suggested direction

[...]

PR #230, opened 2026-09-20
fix(gke): resolve version aliases on UpdateNodePool nodeVersion

### LINKED ISSUE #229: [BUG] GKE UpdateNodePool stores nodeVersion verbatim, so a pool reports "latest" or "-" as its version
## Service

GKE (Kubernetes Engine)

## GCP API / Method

`container.v1.ClusterManager.UpdateNodePool`, the `node_version` field.

## Expected behavior

`node_version` accepts version aliases as well as explicit versions. [cluster_service.proto:4317-4325](https://github.com/googleapis/googleapis/blob/aa87617d67/google/container/v1/cluster_service.proto#L4317):

> Users may specify either explicit versions offered by Kubernetes Engine or version aliases, which have the following behavior:
> - "latest": picks the highest valid Kubernetes version
> - "1.X": picks the highest valid patch+gke.N patch in the 1.X version
> - "1.X.Y": picks the highest valid gke.N patch in the 1.X.Y version
> - "1.X.Y-gke.N": picks an explicit Kubernetes version
> - "-": picks the Kubernetes master version

So after an upgrade the pool should report a resolved version, never the alias.

## Actual behavior

`GkeService.updateNodePool` stores the value as sent:

```java
String nodeVersion = (String) body.get("nodeVersion");
if (nodeVersion != null) {
    pool.setVersion(nodeVersion);
}
```

`GetNodePool` then reports `latest` or `-` as if it were a version. The resolver that handles exactly this, `resolveNodeVersion`, already exists in the same class and is used by `UpdateCluster` and `UpdateMaster`. This is the same defect #197 fixed for `masterVersion` and #198 for `desiredNodeVersion`, on the one call site that was missed.

The raw `(String)` cast is a second, smaller problem: a non-string `nodeVersion` raises an unmapped `ClassCastException` and a 500, where GCP returns 400 `INVALID_ARGUMENT`. Reading it through the existing `stringField` helper fixes both at once.

## Reproduction

```bash
# create a cluster and a node pool, then upgrade the pool to an alias
curl -sX POST "localhost:4588/container/v1/projects/p/locations/us-central1/clusters/c/nodePools/np" \
  -H 'content-type: application/json' -d '{"nodeVersion": "latest"}'
curl -s "localhost:4588/container/v1/projects/p/locations/us-central1/clusters/c/nodePools/np" \
  | python3 -c 'import json,sys; print("version:", json.load(sys.stdin).get("version"))'
# prints: version: latest      expected: the resolved version, floci's DEFAULT_MASTER_VERSION, 1.30.5-gke.1014001
```

## Suggested direction

Route the value through `resolveNodeVersion(field, requested, cluster)` and read it with `stringField`. A test alongside the existing `updateClusterResolvesDesiredNodeVersionAliasesAgainstTheMaster` covering `-`, `latest`, an unmatched `1.X` kept verbatim, and a rejected bare `1` would pin it.

## Environment

- floci-gcp version: `main` at b5a941a
- Found by an audit of the #96 range, not by a user report

avison9:
## Summary

`ClusterManager.UpdateNodePool` stored `nodeVersion` verbatim, so a pool upgraded with `latest` or `-` reported that alias as its version from `GetNodePool`, `ListNodePools` and the cluster's `nodePools[].version`. `node_version` documents the same five spellings as `desired_node_version`, and the resolver for them already exists in `GkeService`; this was the one call site the #197 / #198 / #199 series missed. The value was also read with a raw `(String)` cast, so `{"nodeVersion": 123}` was a 500 with a `ClassCastException` instead of GCP's 400.

Closes #229

```bash
# BEFORE
curl -sX PUT ".../clusters/c/nodePools/default-pool" -H 'content-type: application/json' -d '{"nodeVersion": "latest"}'
curl -s  ".../clusters/c/nodePools/default-pool" | python3 -c 'import json,sys; print(json.load(sys.stdin)["version"])'
# latest                       <- an alias reported as the version
curl -sX PUT ".../clusters/c/nodePools/default-pool" -H 'content-type: application/json' -d '{"nodeVersion": 123}'
# 500, java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.String

# AFTER
# 1.30.5-gke.1014001           <- "latest" resolves to the advertised default
# with "-": 1.29.0-gke.1       <- the cluster's own master version
# with 123: 400 {"error":{"code":400,"message":"nodeVersion must be a string","status":"INVALID_ARGUMENT"}}
```

(The endpoint is `PUT`; the `POST` in the issue's reproduction gets a 405.)

**Root cause:** `updateNodePool` assigned `body.get("nodeVersion")` straight to the pool instead of routing it through `resolveNodeVersion`, which the other three version fields already use.

**What changed** (`GkeService.updateNodePool`):
- Loads the cluster, since `-` on a node field means the cluster's master version.
- Rejects a present non-string `nodeVersion` with 400 `INVALID_ARGUMENT` (same treatment as `"Each entry in 'nodePools' must be an object"` elsewhere in the class), reads it through `stringField` (absent or blank stays a no-op, the field is optional here), and resolves it with `resolveNodeVersion("nodeVersion", ...)` so a rejection names the field the client sent.
- Resolves before mutating, so a rejected value leaves version and etag untouched.
- `resolveMasterVersion` javadoc: "three callers" is now four.

One deviation from the suggested direction: `stringField` on its own would turn `{"nodeVersion": 123}` into a silent 200 no-op rather than the 400 the issue names, so there is an explicit type check in front of it.

Deliberately not asserted in the new tests: the returned `operationType`, so #228 stays a pure rename of `UPDATE_NODE_POOL`.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

Reference: `cluster_service.proto` `UpdateNodePoolRequest.node_version` at googleapis `aa87617d67`, which documents `latest`, `1.X`, `1.X.Y`, `1.X.Y-gke.N` and `-` ("picks the Kubernetes master version"). Before: the alias was stored and reported. After: it resolves exactly as `UpdateCluster.desiredNodeVersion` already does, and a non-string value is a 400 in the GCP error shape. Verified with the Quarkus REST integration test through the real controller; the compatibility suites were not run locally (Docker matrix), and none of them exercises `UpdateNodePool`.

## Checklist

- [x] `./mvnw test` passes locally: 1099/0/0/0 on `main` (97ea835), 1103/0/0/0 with this change
- [x] New or updated integration test added: `GkeUpdateNodePoolRestIntegrationTest` (2 tests) plus `GkeServiceTest.updateNodePoolResolvesNodeVersionAliasesAgainstTheMaster` and `updateNodePoolRejectsInvalidVersionsWithoutTouchingThePool`. All four fail on `main` (`expected: <1.30.5-gke.1014001> but was: <latest>`, and the 500) and pass with the fix.

Two related pre-existing points, left out on purpose, in case you want issues for them: `updateCluster` still reads `desiredNodeVersion` / `desiredMasterVersion` with a raw `(String)` cast (same 500 for a non-string), and `UpdateNodePool` does not move the cluster's `currentNodeVersion` aggregate the way `UpdateCluster.desiredNodeVersion` does.


hectorvent:
Thank you for turning this around so fast, and for the before and after transcripts; they made the change easy to follow.

I checked the alias rules against the [proto](https://github.com/googleapis/googleapis/blob/aa87617d67/google/container/v1/cluster_service.proto#L4317): `-` on a node field means the cluster's master version rather than the server default, and that is exactly what the resolver's `-` branch returns, so the new test asserting `1.29.0-gke.1` is pinning the right thing. Resolving before mutating is a nice touch too.

I also reverted `GkeService` to main inside a worktree and re-ran your two service cases: both fail there with the message you quoted, and the GKE suite is 62 passing on the merge result.

Good catch on the two pre-existing points in your closing note. I will open issues for those rather than grow this PR.

No blockers from my side.

hectorvent (approved):
Approving: alias rules verified against the proto, tests confirmed failing without the fix, GKE suite 62/0 on the merge result.

PR #234, opened 2026-09-20
fix(gke): report only container.v1 Operation.Type members as operationType

### LINKED ISSUE #228: [BUG] GKE returns 11 operationType values that container.v1 Operation.Type does not declare

avison9:
## Summary

`OperationType` had 23 constants, 11 of which are not members of `container.v1.Operation.Type` (a closed 19-value enum, cluster_service.proto at googleapis `aa87617d67`). Ten RPCs emitted them verbatim, so a generated client reading `operationType` for `SetLocations`, `SetNodePoolAutoscaling`, `UpdateNodePool`, IP rotation and the other cluster-level `set*` calls got a value it cannot represent: the Java HttpJson stub decodes it as `TYPE_UNSPECIFIED`.

Closes #228

```bash
# BEFORE
curl -sX POST ".../clusters/c:setLocations" -H 'content-type: application/json' -d '{"locations":["us-central1-a"]}' \
  | python3 -c 'import json,sys; print(json.load(sys.stdin)["operationType"])'
# SET_LOCATIONS      <- not in Operation.Type; google-cloud-container reads TYPE_UNSPECIFIED

# AFTER
# UPDATE_CLUSTER     <- what real GKE reports for the cluster-level set* RPCs

# the issue's comm check (floci constants minus the proto set): 11 names before, none after
```

**Root cause:** the enum was written one constant per RPC, named after the RPC, rather than held to the proto's value set; real GKE folds most of those RPCs into `UPDATE_CLUSTER` and `UPGRADE_NODES`.

**What changed**, following the direction in the issue:
- `OperationType` keeps the 12 members that were already real plus `UPGRADE_NODES`; the 11 invented names are removed.
- `GkeService`: `updateNodePool` and `rollbackNodePoolUpgrade` report `UPGRADE_NODES`; `setAddonsConfig`, `setLoggingService`, `setMonitoringService`, `setLocations`, `setLegacyAbac`, `startIpRotation`, `completeIpRotation` and `setNodePoolAutoscaling` report `UPDATE_CLUSTER`. `COMPLETE_NODE_POOL_UPGRADE` had no call site (that RPC returns `Empty`).
- Persistence: `GkeOperationService` stores operations as `gke-operations.json` through Jackson, and `PersistentStorage` quarantines a file it cannot deserialize, so removing the names outright would drop every persisted GKE operation on the first start after upgrading. A `@JsonCreator` on `OperationType` maps each retired name to the member its RPC reports now; unknown names still fail as before.

On the open question in the issue: the HttpJson stub does not throw on an unknown value, it yields `TYPE_UNSPECIFIED` (protobuf `JsonFormat` with unknown fields ignored), so the new SDK assertions would have failed on `TYPE_UNSPECIFIED` rather than on a parse error.

Overlap note: #230 edits the body of `updateNodePool` just above the return this PR changes; whichever lands second needs a trivial rebase.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

Reference: `Operation.Type` in cluster_service.proto at `aa87617d67` L3984-L4107. Before: ten RPCs returned strings outside that enum. After: every value the emulator can write is a member, verified by a parameterized test over `OperationType.values()`, and decoded through `google-cloud-container`'s `ClusterManagerClient` in `sdk-test-java` (`setNodePoolSize` -> `SET_NODE_POOL_SIZE`, `setLocations` -> `UPDATE_CLUSTER`). The compat suite was compiled locally but not run outside CI.

## Checklist

- [x] `./mvnw test` passes locally: 1099/0/0/0 on `main` (97ea835), 1124/0/0/0 with this change
- [x] New or updated integration test added: `OperationTypeTest` (25 cases: proto-set membership for every constant, the 11 retired names loading from the persisted `Map<String, StoredOperation>` shape, round trip and unknown-name failure), four `GkeServiceTest` assertions moved to the real members, and two new SDK cases in `compatibility-tests/sdk-test-java/GkeTest`


hectorvent:
Thank you, this closes the issue exactly, and the compatibility cases are the part that matters most: asserting through `Operation.Type` on the generated client is what proves the decode, which the unit tests could never show.

I checked the enum mechanically against the proto rather than by eye: all 13 remaining constants are members of the 19 value [`Operation.Type`](https://github.com/googleapis/googleapis/blob/aa87617d67/google/container/v1/cluster_service.proto#L3984), with nothing left outside it. `UPDATE_CLUSTER` for the cluster level `set*` calls and `UPGRADE_NODES` for the node version paths both read correctly against those members' own descriptions.

The `RETIRED` map is a nice catch that I missed when filing this: without it a persisted operations file holding one of the old names would fail to load and take every stored operation with it.

One open question, no change needed: `SetNodePoolAutoscaling` maps to `UPDATE_CLUSTER` because the enum has no autoscaling member. That seems like the only sensible fallback, though if you ever have a real capture handy it would be worth confirming.

No blockers from my side.

hectorvent (approved):
Approving: enum verified mechanically against the proto's 19 members, Gke*+OperationTypeTest 87/0 locally.

PR #236, opened 2026-09-20
fix(gke): reject non-string UpdateCluster version fields with 400

### LINKED ISSUE #232: [BUG] GKE UpdateCluster returns 500 for a non-string desiredNodeVersion or desiredMasterVersion
## Service

GKE (Kubernetes Engine)

## GCP API / Method

`container.v1.ClusterManager.UpdateCluster`, the `desiredNodeVersion` and `desiredMasterVersion` fields.

## Expected behavior

A well-formed JSON request carrying a value of the wrong type is a client error. Everywhere else in this service the repo returns 400 in the standard GCP error body, for example `{"error": {"code": 400, "message": "...", "status": "INVALID_ARGUMENT"}}`, which is what `GcpExceptionMapper` produces from a `GcpException`.

#230 established exactly this for `UpdateNodePool.nodeVersion`: a non-string value now returns 400 `INVALID_ARGUMENT` rather than a 500.

## Actual behavior

`GkeService.updateCluster` still reads both fields with a raw cast:

```java
String desiredNodeVersion = (String) updateMap.get("desiredNodeVersion");   // GkeService.java:328
String desiredMasterVersion = (String) updateMap.get("desiredMasterVersion"); // GkeService.java:346
```

`{"update": {"desiredMasterVersion": 123}}` therefore raises an unmapped `ClassCastException`, which surfaces as a 500 with no GCP error body at all.

## Reproduction

```bash
curl -sX PUT ".../clusters/c" -H 'content-type: application/json' \
  -d '{"update": {"desiredMasterVersion": 123}}'
# 500, java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.String
# expected: 400 {"error":{"code":400,"message":"desiredMasterVersion must be a string","status":"INVALID_ARGUMENT"}}
```

## Suggested direction

Mirror what #230 did for `nodeVersion`: an explicit type check in front of `stringField`, so an absent field stays the no-op it is today and a wrong-typed one is a 400 naming the field.

Worth knowing before picking this up: a wider audit counted the same pattern at roughly ten call sites in `GkeService` and three in `KubernetesController`, so a single helper (for example a `mapField` sibling to `stringField`) would cover the rest in one pass. The two fields above are the ones with a known user-visible path, since Terraform sends `desiredMasterVersion` on cluster updates.

## Environment

- Reported by @avison9 in the closing note of #230

avison9:
## Summary

`UpdateCluster` read `desiredNodeVersion` and `desiredMasterVersion` with a raw `(String)` cast, so `{"update": {"desiredMasterVersion": 123}}` escaped as an unmapped `ClassCastException`: a 500 with a stack trace and no GCP error body. Terraform sends `desiredMasterVersion` on cluster updates. This mirrors what #230 did for `UpdateNodePool.nodeVersion`, through one shared helper.

Closes #232

```bash
# BEFORE
curl -sX PUT ".../clusters/c" -H 'content-type: application/json' -d '{"update": {"desiredMasterVersion": 123}}'
# 500  java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.String
#      at io.floci.gcp.services.gke.GkeService.updateCluster(GkeService.java:346)

# AFTER
# 400  {"error":{"code":400,"message":"desiredMasterVersion must be a string","status":"INVALID_ARGUMENT"}}
```

**Root cause:** the update paths used a raw cast while the create paths used the type-safe `stringField`; Jackson delivers a JSON number as an `Integer`, and the cast throws outside any `GcpException` mapping.

**What changed** (`GkeService`):
- `optionalStringField(map, field)`: absent or blank is `null` (proto3 JSON treats `""` as unset), any other type is a 400 naming the field. A sibling of `stringField`, as suggested in the issue.
- `updateCluster` reads `desiredNodeVersion` and `desiredMasterVersion` through it, before any mutation; `nodeVersionUpdateTargets` reads `desiredNodePoolId` through it too, since it is on the same `desiredNodeVersion` path Terraform exercises (`desiredNodePoolId: "default-pool"`) and had the same 500.
- `updateNodePool`'s inline check from #230 is replaced by the helper, so the three sites share one implementation and one message shape.

One behavior detail: a blank `desiredMasterVersion` / `desiredNodeVersion` used to reach the resolver and fail as `Invalid desiredMasterVersion ""`; it is now a no-op, consistent with proto3 and with `UpdateNodePool`. No SDK or gcloud path sends `""`.

Scope: the wider sweep the issue mentions (about ten more casts in `GkeService`, three root-object casts in `KubernetesController`) is left for its own pass as suggested. With the helper in place it is mechanical: `(String) map.get(x)` becomes `optionalStringField(map, x)` wherever the field is optional.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

Before: three `ClusterUpdate` fields returned a 500 for a wrongly typed value. After: 400 `INVALID_ARGUMENT` in the standard error body, which is what GCP's front end returns for a type mismatch on a JSON field. Verified through the Quarkus REST integration test against the real controller; the compatibility suites do not send malformed bodies, so none was changed.

## Checklist

- [x] `./mvnw test` passes locally: 1126/0/0/0 with this change on `main` (8009e11); the two new tests are the only delta, every other class is unchanged
- [x] New or updated integration test added: `GkeUpdateClusterRestIntegrationTest.aNonStringVersionFieldIsABadRequestNotAServerError` and `GkeServiceTest.updateClusterRejectsNonStringVersionFieldsWithoutTouchingTheCluster`; both fail on `main` with the `ClassCastException` at `GkeService.java:346` and pass here


avison9:
Good catch, and a real one: with `{"desiredNodeVersion": "latest", "desiredMasterVersion": 123}` the node version (and the pool etag) moved before the master field was read, so the 400 left a partial update. Fixed in the second commit: `updateCluster` now reads both fields through `optionalStringField`, resolves the node pool targets and both versions, and only then touches the cluster or the pools. Both tests gained that case (plus `desiredMasterVersion: "1"`, a string the resolver rejects, beside a valid node version), with the cluster created on `1.29.0-gke.1` so an applied `latest` is observable; they fail on the previous commit (`expected: <1.29.0-gke.1> but was: <1.30.5-gke.1014001>`) and pass now. Full suite 1160/0/0/0 on the rebased branch.

hectorvent:
Thanks, and sorry about the collision: #235 opened on the same issue two minutes before this one, because the issue had no assignee to signal it was taken. That is our process gap, not yours, and I am sorting out the ordering rather than leaving you both guessing.

Resolving both versions before the first assignment is the right fix, and the case you added, a valid `desiredNodeVersion` beside a master value the resolver rejects, is exactly the one that used to half apply. Asserting the pool etag as well as its version is a good touch. The helper itself is the shape the issue asked for, and treating `""` as unset matches proto3 JSON.

**(blocking)** `desiredLoggingService` and `desiredMonitoringService`, a few lines below, still read through raw casts, so a 500 is still reachable from the same method. `optionalStringField` covers them in a line each.

Nothing else from my side.

avison9:
Thanks, and no worries about the ordering; whichever of #235 and this one you land, the other rebases in minutes. Since #235 keeps the blank-string rejection and this one treats `""` as unset per proto3 JSON, that is the one behavioural difference to pick between.

The blocking item is fixed in the third commit: `desiredLoggingService` and `desiredMonitoringService` now go through `optionalStringField`, and both are read up front with the version fields, before anything is applied, so a bad one cannot leave a resolved version behind either. The unit and REST tests each gained a case for them (`{"desiredLoggingService": 123}` and a non-string `desiredMonitoringService` beside a valid `desiredNodeVersion`, with the cluster asserted untouched). Full suite 1160/0/0/0.

hectorvent:
Thanks, that closes the last point: both service fields now read through the same helper, so no raw cast is left in the method.

Two things, neither about your code:

#238 landed a few minutes ago and touches the same method, so this branch conflicts in `GkeService.java` now. A rebase should be quick, and the conflict is mechanical.

On the overlap, one correction to your note above: #235 pushed a fix three minutes before you wrote it, so it now treats a blank the same way this does. The remaining difference is scope, since this one also routes `desiredNodePoolId`, the two service fields and the check #230 added through the helper, and resolves both versions before mutating.

No review blockers from my side; I will confirm the order shortly so neither of you is waiting on it.

hectorvent:
Order confirmed, and thank you for offering to go either way.

#235 goes in first: it was opened a couple of minutes earlier, it fixes the issue as filed, and it is a first contribution that has been through three rounds. Then this one rebases on top.

Everything this adds beyond that survives the rebase and is the reason it is worth landing second: the shared helper, `desiredNodePoolId` and the two service fields routed through it, the check from #230 folded in, and resolving both versions before any mutation, which also closes the half apply that Greptile spotted.

I will watch for the rebase on both and merge in that order.

avison9:
Thanks for settling the order, and #235 first makes sense. Rebased onto main with #238 in: the one conflict was the aggregate write in `updateCluster`, resolved by keeping this branch's read-and-resolve-both-first ordering and letting #238's minimum-over-pools computation own the `currentNodeVersion` write after the pool loop. Three commits as before, full suite 1178/0/0/0, `Gke*` 70/0. I will rebase again once #235 lands.

hectorvent (approved):
Thanks, and thanks for taking the ordering in good part.

I checked the rebase rather than assuming: #238's minimum-over-pools write is the unmodified line in the merged result, exactly as you described, and its test is untouched. Your own coverage is all present and CI is green across all twelve checks.

One correction to something I said last round. I described this as a refactor of #235, and that undersold it: hoisting resolution above the mutation block also closes a case where a 400 could leave a partial update behind, which #235 does not fix. That is worth keeping, so the plan stands: #235 first, then this rebased on top.

Nothing else from my side.

hectorvent:
#235 is merged now, and #239 went in too, so this one is unblocked. It conflicts in `GkeService` and `GkeUpdateClusterRestIntegrationTest`, which is exactly what we expected given both changes touch the same block.

Would you rebase onto main when you get a chance? The approval is already on the PR, so once CI is green again there is nothing further from my side.

Thanks again for taking the ordering the way you did.

avison9:
Rebased onto main with #235 and #239 in, and squashed to a single commit while I was at it, since the three had become one change.

Both conflicts were the expected ones:

- `GkeService.updateCluster`: kept this branch's version, which reads every field through `optionalStringField` and resolves both versions before the first setter. It replaces #235's inline `instanceof` checks with the helper, and the error text is unchanged (`<field> must be a string`), so #235's assertions still hold.
- `GkeUpdateClusterRestIntegrationTest`: kept both methods side by side, #235's `nonStringVersionsAreBadRequestsAndLeaveTheClusterUntouched` and this PR's `aNonStringVersionFieldIsABadRequestNotAServerError`.

Nothing in the diff uses `var` or a wildcard import, per the new AGENTS.md rules. Full suite 1278/0/0/0 on `db6dc93`, `Gke*` 99/0.

hectorvent:
Thanks for the rebase and the squash, both look right. I checked the result: #235's inline checks now go through `optionalStringField`, its test still passes unchanged next to yours, and every rejection runs before the first setter, so a 400 leaves the cluster and its pools untouched.

On the bot note about empty `desiredLoggingService` and `desiredMonitoringService`: both are plain proto3 `string` fields with no presence ([cluster_service.proto](https://github.com/googleapis/googleapis/blob/aa87617d6732d348ff158d20d474c0dfba3d0086/google/container/v1/cluster_service.proto#L3631)), so an explicit `""` cannot be told apart from an absent field. Treating blank as unset is the right call, nothing to change there.

The approval stands and CI is green, so no blockers from my side.

Issue #216, opened 2026-09-16
[FEAT] Managed Kafka: implement the Kafka Connect service (connectClusters, connectors)


avison9:
## Service

Managed Kafka (Kafka Connect)

## API Method / Feature

`google.cloud.managedkafka.v1.ManagedKafkaConnect`, the whole service: `ListConnectClusters`, `GetConnectCluster`, `CreateConnectCluster`, `UpdateConnectCluster`, `DeleteConnectCluster`, `ListConnectors`, `GetConnector`, `CreateConnector`, `UpdateConnector`, `DeleteConnector`, `PauseConnector`, `ResumeConnector`, `RestartConnector`, `StopConnector`, under `projects/*/locations/*/connectClusters/*` and `.../connectors/*`.

## GCP Documentation

- `managed_kafka_connect.proto`: https://github.com/googleapis/googleapis/blob/master/google/cloud/managedkafka/v1/managed_kafka_connect.proto
- REST reference: https://cloud.google.com/managed-service-for-apache-kafka/docs/reference/rest/v1/projects.locations.connectClusters
- Concepts: https://cloud.google.com/managed-service-for-apache-kafka/docs/connect-cluster/overview

## Why is this needed?

Managed Kafka's Connect surface is not served at all; `ConnectCluster` and `Connector` calls are 404s. Pipelines that pair a Kafka cluster with a Connect cluster (sink to GCS or BigQuery, MirrorMaker between clusters) cannot be exercised locally, and Terraform `google_managed_kafka_connect_cluster` / `google_managed_kafka_connector` have no target.

## Proposed scope

Phased, so the control plane can land without the sidecar:

- **Phase 1, control plane.** `ConnectCluster` (references a `kafka_cluster` resource name, `capacity_config`, `gce_config`, state `CREATING`/`ACTIVE`) and `Connector` (name, `configs` map, `task_restart_policy`, output-only `state`) CRUD, plus the four lifecycle custom methods flipping `state` between `RUNNING`, `PAUSED`, `STOPPED`. Immediately-done LROs for cluster create/update/delete, as the Kafka cluster RPCs do today. Same-project / same-location validation of the referenced Kafka cluster.
[...]

Issue #218, opened 2026-09-16
[FEAT] Managed Kafka: asynchronous cluster operations (CREATING state, pollable LRO) instead of synchronous done=true


avison9:
## Service

Managed Kafka (and, by the same pattern, the other Docker-backed services)

## API Method / Feature

`ManagedKafka.CreateCluster` / `UpdateCluster` / `DeleteCluster` long-running operations and `google.longrunning.Operations.GetOperation` / `ListOperations` under `projects/*/locations/*/operations/*`; `Cluster.state` lifecycle (`CREATING` → `ACTIVE`, `DELETING`).

## GCP Documentation

- `managed_kafka.proto` (`returns (google.longrunning.Operation)` on the cluster RPCs, `OperationMetadata`): https://github.com/googleapis/googleapis/blob/master/google/cloud/managedkafka/v1/managed_kafka.proto
- `Cluster.State`: https://github.com/googleapis/googleapis/blob/master/google/cloud/managedkafka/v1/resources.proto
- LRO polling contract: https://cloud.google.com/managed-service-for-apache-kafka/docs/reference/rest/v1/projects.locations.operations

## Why is this needed?

Today `CreateCluster` blocks until the Redpanda container is ready and then returns an operation with `done: true` and the cluster already `ACTIVE`; `UpdateCluster` and `DeleteCluster` do the same. Real Managed Kafka returns a `RUNNING` operation within a second, with the cluster in `CREATING`, and the client polls `operations.get` (the SDK's `createClusterAsync(...).get()` does exactly this; `gcloud managed-kafka clusters create` polls too).

Two consequences of the synchronous model:

1. Client-side timeouts. A cold Redpanda start can take longer than the Google HTTP client's default read timeout, and the client then has no operation name to poll because none was returned. The same effect was measured on Cloud SQL for MySQL in #202 (15 to 25 s cold start against a 20 s default), where it is documented as a limitation.
2. Code paths that handle `CREATING` (retry loops, readiness gates, Terraform's `PENDING` handling) are never exercised locally, so a client bug in that path ships untested.

The current shape is a deliberate emulator-wide choice (the GKE docs say so explicitly: "all mutations return a synchronous, DONE Operation"), so this is a design question for the maintainer before it is a work item.

## Proposed scope, if accepted

[...]

PR #237, opened 2026-09-20
feat(kafka): implement the ManagedKafkaConnect control plane

avison9:
## Summary

Implements the control plane of `google.cloud.managedkafka.v1.ManagedKafkaConnect`: `ConnectCluster` and `Connector` CRUD, the four connector lifecycle methods, done LROs on cluster mutations, and same-project/location validation of the referenced Kafka cluster. This is Phase 1 of #216 as split in the issue; the Connect runtime sidecar (Phase 2) is a follow-up that keeps this surface.

Part of #216 (Phase 1)

```bash
# BEFORE
B=http://localhost:4588/v1/projects/p/locations/us-central1
curl -s -o /dev/null -w '%{http_code}\n' -X POST "$B/connectClusters?connectClusterId=cc" -H 'Content-Type: application/json' -d '{...}'
# 404

# AFTER
# {"name":".../operations/<uuid>","done":true,"response":{"name":".../connectClusters/cc","kafkaCluster":".../clusters/c",
#   "capacityConfig":{"vcpuCount":12,"memoryBytes":21474836480},"gcpConfig":{...},"state":"ACTIVE",...}}
curl -s -X POST "$B/connectClusters/cc/connectors?connectorId=gcs-sink" -H 'Content-Type: application/json' \
  -d '{"configs":{"connector.class":"io.aiven.kafka.connect.gcs.GcsSinkConnector","topics":"orders"}}'
# {"name":".../connectors/gcs-sink","configs":{...},"state":"RUNNING"}
curl -s -X POST "$B/connectClusters/cc/connectors/gcs-sink:pause" -H 'Content-Type: application/json' -d '{}'
# {}    -> GET reports "state":"PAUSED"
```

### What changed

- **`KafkaConnectService`, `KafkaConnectController`, `StoredConnectCluster`, `StoredConnector`, `ConnectorState`** (new). The 14 RPCs from `managed_kafka_connect.proto` on the same `/v1/projects/{p}/locations/{l}` root as `KafkaController`. Cluster mutations return the done-LRO envelope the Kafka cluster RPCs use; `DeleteConnector` returns `{}` (`Empty`); `:pause` / `:resume` / `:restart` / `:stop` return their empty response messages and move `state` to `PAUSED` / `RUNNING` / `RUNNING` / `STOPPED`.
- **Validation from the proto:** `kafkaCluster` REQUIRED, a `projects/{p}/locations/{l}/clusters/{c}` name in the same project and location (400) that exists (404); `capacityConfig` REQUIRED with positive `vcpuCount` and `memoryBytes` (the proto3 JSON string form of int64 accepted); `gcpConfig.accessConfig.networkConfigs` REQUIRED with a `primarySubnet` each; `labels` / `config` / `configs` as string maps; `taskRestartPolicy` durations as `"60s"`. `kafkaCluster` is IMMUTABLE: naming it in `updateMask` or changing it is a 400.
- **`updateMask`** honored when sent (camelCase, or gcloud's snake_case); without one, the mutable fields present in the body apply, the same leniency `UpdateCluster` has today. Lists are name-sorted and paged with `PageToken`. Deleting a Connect cluster deletes its connectors.
- **`KafkaService`:** two lines: `KafkaConnectController` registered on the `kafka` descriptor (so `FLOCI_GCP_SERVICES_KAFKA_ENABLED` gates it) and a package-private `clusterExists`. Kept as separate classes so this does not collide with #219, which edits `KafkaController` and `KafkaService` elsewhere.
- **Docs:** Kafka Connect section in `managed-kafka.md` with the limitation stated (no runtime, connectors do not move data yet), and the 14 RPCs in the operations list.
- **Compat:** `KafkaTest.connectClusterAndConnectorLifecycle` in `sdk-test-java`, raw HTTP like the rest of that class.

A Connect cluster is `ACTIVE` at create in both mock and Docker mode: nothing is provisioned in this phase, so there is no `CREATING` window to report and no `kafka.mock` gate to add. Phase 2 will gate the sidecar on the existing flag as AGENTS.md prescribes.

Left for follow-ups, stated so they are not mistaken for oversights: deleting a Kafka cluster still referenced by a Connect cluster is allowed (the real service likely refuses with `FAILED_PRECONDITION`); `request_id`, `filter` and `order_by` are accepted and ignored as elsewhere; connector `configs` are not validated against a Connect runtime.

### Concurrency and storage invariant (CONTRIBUTING.md "Concurrency and Storage Invariants")

- **Invariant:** a connector exists only under a Connect cluster that exists, and no connector write is lost to a concurrent write on the same connector.
- **Competing paths:** `DeleteConnectCluster` (cascade) against `CreateConnector`, `UpdateConnector`, `DeleteConnector`, the four lifecycle methods, `GetConnector`, `ListConnectors`, `GetConnectCluster`, `UpdateConnectCluster`; and any two mutations of one connector against each other.
- **Serialization:** every operation on a Connect cluster or its connectors runs under that cluster's lock, striped on the cluster name (`KafkaConnectService.clusterLock`, 256 stripes, the same shape as `SecretManagerService`). Request bodies are validated before the lock is taken, so a rejected body never mutates the stored object. `DeleteConnectCluster` removes connectors first and the cluster last, all under the lock.
- **Lock order:** this service takes exactly one lock per operation, so there is no order to preserve and no path acquires two.
- **Deferred, stated:** a Kafka `DeleteCluster` racing `CreateConnectCluster` can leave a Connect cluster whose `kafkaCluster` no longer exists. The reference is checked at create and never dereferenced afterwards in this phase, so no operation can fail or orphan state because of it; refusing Kafka cluster deletion while a Connect cluster references it (what the real service does) is listed as a follow-up and belongs with Phase 2, when the reference starts to matter.
- **Test:** `KafkaConnectServiceTest.deleteSerializesConnectorMutationsAndLeavesNoOrphan` blocks the delete inside the connector store's first `delete` (holding the lock), submits eight competing operations, asserts none completes while the delete is open (deterministic latch, no sleeps), then releases and asserts every one of them observes `NOT_FOUND`, the connector store is empty (the connector submitted during the window was never stored), and the cluster name is reusable.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

Shapes and field behaviors from `managed_kafka_connect.proto` and `resources.proto` (googleapis master): `ConnectCluster` (`kafka_cluster` REQUIRED IMMUTABLE, `capacity_config` REQUIRED, `gcp_config.access_config.network_configs` REQUIRED with `primary_subnet`), `Connector` (`configs`, `state` OUTPUT_ONLY, `task_restart_policy` oneof), HTTP bindings (`body: "connect_cluster"` / `"connector"`, `body: "*"` on the lifecycle methods, `Empty` on `DeleteConnector`). Verified with the Quarkus REST test through the real controllers and by running the `sdk-test-java` `KafkaTest` (12/12) against the packaged emulator in Kafka mock mode. The suite has no `google-cloud-managedkafka` dependency, so the compat case is raw HTTP like the existing Kafka cases.

## Checklist

- [x] `./mvnw test` passes locally: 1133/0/0/0 across 131 classes (129 classes / 1124 tests on `main` at 8009e11 by derivation; the 9 new tests are the only delta)
- [x] New or updated integration test added: `KafkaConnectRestIntegrationTest` (Quarkus, end to end), `KafkaConnectServiceTest` (8 unit tests, one a deterministic interleaving test), and `KafkaTest.connectClusterAndConnectorLifecycle` in `compatibility-tests/sdk-test-java`


hectorvent:
Thank you, this is a big surface and it reads cleanly.

I enumerated the service rather than spot checking: all 14 ManagedKafkaConnect RPCs are here, and each route matches its http binding, including the four colon suffixed connector methods. `Connector.State` matches the [proto's seven values](https://github.com/googleapis/googleapis/blob/aa87617d67/google/cloud/managedkafka/v1/resources.proto#L462) exactly, and returning `{}` from pause, resume, restart and stop is right, since those response messages are empty in the proto.

The connectCluster create, update and delete calls return the same `done: true` envelope the existing Kafka cluster RPCs use, so they stay consistent with the rest of the service; making those genuinely asynchronous is already tracked separately in #218.

Tests pass locally alongside the existing Kafka suites.

No blockers from my side.

avison9:
All four findings addressed in 24bbc82; each one was real.

1. **Nested masks:** `FieldMask` (a small record in `KafkaConnectService`) now keeps the full paths. A nested path such as `capacityConfig.vcpuCount`, `labels.env` or `taskRestartPolicy.minimumBackoff` replaces only that leaf in a copy of the stored value (or clears it when the body does not carry it) and the merged value is validated as a whole, so `capacityConfig.vcpuCount` with only `vcpuCount` in the body keeps `memoryBytes`, and `taskRestartPolicy.minimumBackoff` keeps `maximumBackoff`. A whole-field path or no mask behaves as before. Pinned by `nestedMaskPathsReplaceOnlyTheNamedLeaf` and, through the SDK, by `task_restart_policy.minimum_backoff` in the compat test.
2. **int64 parsing:** `requirePositiveLong` goes through `BigDecimal`: a JSON number must be integral (`12.5` is a 400, `12.0` is 12), a string must be a plain decimal as proto3 JSON spells int64 (`"1e3"`, `"12.5"`, `"abc"` are 400s), anything above `Long.MAX_VALUE` is a 400, and `NumberFormatException` never escapes. `int64FieldsMustBeWholeNumbersInRange` covers the cases including `Long.MAX_VALUE` accepted as a string.
3. **Resource ids:** `connectClusterId` and `connectorId` must match `[a-z]([-a-z0-9]{0,61}[a-z0-9])?`, the grammar the API documents, so an encoded slash or a colon is a 400 instead of a stored resource the routes cannot reach. `resourceIdsFollowTheDocumentedGrammar`. (The Kafka `clusterId` on `CreateCluster` has never been validated; left alone here as pre-existing.)
4. **SDK coverage:** the raw-HTTP compat case is replaced by `KafkaConnectTest` in `sdk-test-java`, driving the whole surface through `ManagedKafkaConnectClient` over HttpJson: `createConnectClusterAsync(...).get()` unpacking the LRO response, get/list, `updateConnectClusterAsync` with a `FieldMask`, connector create/update/delete, the four custom methods, and `deleteConnectClusterAsync`. Two things this surfaced and fixed: the LRO `response` now carries its `Any` type URL (`type.googleapis.com/google.cloud.managedkafka.v1.ConnectCluster`, `google.protobuf.Empty` on delete), which the generated client needs; and `libraries-bom` 26.53.0 pins `google-cloud-managedkafka` 0.12.0, which predates the Connect client, so `google-cloud-managedkafka-bom` 0.17.0 (the first release with it) is imported ahead of it, moving the three managedkafka artifacts together. Ran locally against the packaged emulator in Kafka mock mode: `KafkaConnectTest` 5/5, `KafkaTest` 12/12 unchanged.

Full suite 1170/0/0/0 on the rebased branch.

hectorvent:
Thanks for the follow-up, this is stronger than what I reviewed first.

The nested mask handling is the part that matters: walking `capacityConfig.vcpuCount` rather than replacing the whole submessage is what the mask contract asks for, and rejecting a path whose top-level field is not mutable is a good guard. Strict int64 bounds and the id checks close the other obvious holes.

Moving the Connect cases into their own `KafkaConnectTest` through `ManagedKafkaConnectClient` is the best part: SDK based coverage proves the decode, which raw HTTP never can, and pinning the managedkafka BOM ahead of libraries-bom with a comment explaining why is the right way to handle that version gap.

One thing for later, not for this PR: `update_mask` is marked required on both update RPCs, and here a missing mask updates whatever the body carries. That matches what Cloud Scheduler already does in this repo, so it is our convention rather than your choice. I will open one issue covering every service.

Tests pass locally alongside the ACL suites. No blockers from my side.

hectorvent (approved):
Approving: 14/14 RPCs and the state enum verified against the proto, nested masks and SDK-based compat added in round 2, Kafka suites 18/0 locally.

PR #238, opened 2026-09-20
fix(gke): recompute currentNodeVersion as the minimum across node pools

### LINKED ISSUE #233: [BUG] GKE UpdateNodePool leaves Cluster.currentNodeVersion stale
## Service

GKE (Kubernetes Engine)

## GCP API / Method

`container.v1.ClusterManager.UpdateNodePool`, and the `Cluster.currentNodeVersion` field it should move.

## Expected behavior

[cluster_service.proto:3086-3091](https://github.com/googleapis/googleapis/blob/aa87617d67/google/container/v1/cluster_service.proto#L3086) defines the field as an aggregate over the node pools:

> Output only. Deprecated, use NodePools.version instead. The current version of the node software components. If they are currently at multiple versions because they're in the process of being upgraded, this reflects the minimum version of all nodes.

So after any node pool's version changes, a `GetCluster` should report the minimum version across that cluster's pools.

## Actual behavior

Only two paths ever write it, and neither is `UpdateNodePool`:

- `createCluster` sets it to the master version (`GkeService.java:243`)
- `updateCluster` with `desiredNodeVersion` sets it to the resolved value (`GkeService.java:336`)

`UpdateNodePool` changes the pool's own `version` and leaves the cluster aggregate untouched, so a cluster whose only pool now runs `1.30.5-gke.1014001` can still report an older `currentNodeVersion`, and one upgraded pool out of several can leave the aggregate above the real minimum.

## Reproduction

```bash
# cluster created at an explicit older version, then its pool upgraded
curl -sX POST ".../clusters" -d '{"cluster": {"name": "c", "initialClusterVersion": "1.29.0-gke.1"}}'
curl -sX PUT  ".../clusters/c/nodePools/default-pool" -d '{"nodeVersion": "latest"}'
curl -s ".../clusters/c" | python3 -c 'import json,sys; d=json.load(sys.stdin); print(d["currentNodeVersion"], [p["version"] for p in d["nodePools"]])'
# 1.29.0-gke.1 ['1.30.5-gke.1014001']    <- aggregate stale, pool upgraded
```

## Suggested direction

Recompute the aggregate from the stored pools whenever a pool's version changes, taking the minimum rather than the last write, and use the same helper in `UpdateCluster` so the two paths cannot drift. A cluster with no pools keeps whatever `createCluster` set.

Comparing versions needs a little care: `1.30.5-gke.1014001` is not ordered correctly by string comparison against `1.9.x`, so the helper wants a small version comparator, which the repo does not have yet.

## Environment

- Reported by @avison9 in the closing note of #230

avison9:
## Summary

`Cluster.currentNodeVersion` is defined by the proto as the minimum version across the cluster's node pools, but floci-gcp wrote it from exactly two places (`CreateCluster`, and `UpdateCluster` with `desiredNodeVersion`, as the last value sent) and never from `UpdateNodePool`. A pool upgraded through `UpdateNodePool` left `GetCluster` reporting a version no pool ran, and upgrading one pool of two moved the aggregate above the pool still behind. The aggregate is now recomputed as the minimum over the stored pools whenever the pool set or a pool's version changes.

Closes #233

```bash
# BEFORE
curl -sX POST ".../clusters" -d '{"cluster": {"name": "c", "initialClusterVersion": "1.29.0-gke.1"}}'
curl -sX PUT  ".../clusters/c/nodePools/default-pool" -d '{"nodeVersion": "latest"}'
curl -s ".../clusters/c" | python3 -c 'import json,sys; d=json.load(sys.stdin); print(d["currentNodeVersion"], [p["version"] for p in d["nodePools"]])'
# 1.29.0-gke.1 ['1.30.5-gke.1014001']    <- aggregate stale

# AFTER
# 1.30.5-gke.1014001 ['1.30.5-gke.1014001']
# two pools at 1.30.0-gke.1, one upgraded to 1.31.5-gke.1: 1.30.0-gke.1 (the minimum), not 1.31.5-gke.1
```

**Root cause:** the field was treated as a value set by the version-changing RPCs instead of a function of the pools, and the repo had no way to compare two GKE versions (`1.9.0-gke.1` sorts after `1.30.5-gke.1014001` as a string).

**What changed:**
- `GkeVersions` (new, package-private): numeric ordering of `MAJOR.MINOR[.PATCH][-gke.N]`, missing parts as 0; anything outside that shape (possible on `initialClusterVersion` and an explicit pool `version`, both stored as sent) sorts last so it never becomes the minimum. `minimum(...)` ignores null and blank.
- `GkeService.refreshCurrentNodeVersion`: minimum over the stored pools, persisted when it changed; a cluster with no pools keeps what it reports.
- Called from every path that changes the pool set or a pool version: `createCluster` (after its initial pools, so `nodePools[].version` below the master is reflected from the start), `createNodePool`, `deleteNodePool`, `updateNodePool`, and `updateCluster` with `desiredNodeVersion`, where the direct assignment is replaced by the minimum (falling back to the resolved version when the cluster has no pools, as before).

The issue names `UpdateNodePool` and `UpdateCluster`; `CreateNodePool` with an older explicit version and `DeleteNodePool` of the oldest pool change the same minimum, so they call the same helper. Two more call sites, no other behaviour.

One deliberate non-change: the cluster etag is not bumped when the aggregate moves. The field is OUTPUT_ONLY and the etag guards user-mutable state, so an `UpdateNodePool` should not invalidate a client's cluster etag.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

Reference: `cluster_service.proto` at googleapis `aa87617d67`, `Cluster.current_node_version` (L3086-L3091): "If they are currently at multiple versions because they're in the process of being upgraded, this reflects the minimum version of all nodes." Before: the field lagged `UpdateNodePool` and took the last written value under `UpdateCluster`. After: the minimum across pools on every pool mutation. One visible change beyond the issue: right after `CreateCluster` with `nodePools[]` carrying explicit versions below the master, `currentNodeVersion` reports their minimum rather than the master version, which is what GKE reports. The compat suites do not read this deprecated field; verified with the Quarkus REST test through the real controller.

## Checklist

- [x] `./mvnw test` passes locally: 1163/0/0/0 across 133 classes (132 classes / 1158 tests on `main` at 4958e13 by derivation; the 5 new tests are the only delta)
- [x] New or updated integration test added: `GkeUpdateNodePoolRestIntegrationTest` extended with the issue's `GetCluster` check; `GkeServiceTest.updateNodePoolMovesTheClusterNodeVersionAggregate` and `currentNodeVersionIsTheMinimumAcrossPools`; `GkeVersionsTest` (3). The three behavioural tests fail on `main` (`expected: <1.30.5-gke.1014001> but was: <1.29.0-gke.1>`) and pass here.


avison9:
Both findings addressed in the second commit.

1. **Overflow:** `GkeVersions.parse` catches `NumberFormatException`, so a component beyond the `long` range (`1.999999999999999999999.0-gke.1`) is an unparseable version: it sorts after every well-formed one and can never be the minimum, and the comparison never throws after a pool has been persisted. `GkeVersionsTest` covers it.
2. **Persisted aggregates:** `init()` now runs `refreshPersistedNodeVersionAggregates()` right after the embedded pool migration: one idempotent pass over the stored clusters, each brought to the minimum over its pools, so a cluster written by an earlier build reads correctly after the upgrade without waiting for a pool mutation. `startupRefreshesAggregatesPersistedByEarlierBuilds` builds such a cluster (aggregate `1.29.0-gke.1`, pools at `1.30.5-gke.1014001` and `1.30.1-gke.7`), calls `init()`, and reads back `1.30.1-gke.7`.

Full suite 1164/0/0/0.

hectorvent:
Thank you, and this is a good example of a fix that goes one level deeper than the issue asked.

I checked the field against the proto, which says the aggregate ["reflects the minimum version of all nodes"](https://github.com/googleapis/googleapis/blob/aa87617d67/google/container/v1/cluster_service.proto#L3086), so deriving it from the stored pools on every path that can change them is right, and extending that to `CreateNodePool` and `DeleteNodePool` beyond the two the issue named is the correct reading.

`GkeVersions` is the part I would have been most worried about, and the numeric ordering with unparseable values sorting last is exactly right, since `initialClusterVersion` and explicit pool versions are stored as sent. Good catch from the bot on overflow, and the idempotent refresh pass for clusters persisted before this change is the thing that would have bitten someone after an upgrade.

Agreed on leaving the cluster etag alone: the field is output only, so a node pool update should not invalidate a client's etag.

No blockers from my side.

hectorvent (approved):
Approving: aggregate semantics verified against the proto, version comparator and the persisted-cluster refresh both checked, Gke* 68/0 locally.

PR #239, opened 2026-09-20
fix(gke): reject a 1.X version alias that matches no valid version

### LINKED ISSUE #231: [BUG] GKE keeps an unmatched 1.X version alias verbatim instead of rejecting it
## Service

GKE (Kubernetes Engine)

## GCP API / Method

`container.v1.ClusterManager`, the `1.X` version alias, on all four version fields (`masterVersion`, `desiredMasterVersion`, `desiredNodeVersion`, `nodeVersion`).

## Expected behavior

[cluster_service.proto:4317-4325](https://github.com/googleapis/googleapis/blob/aa87617d67/google/container/v1/cluster_service.proto#L4317) documents the alias as picking a real version:

> - "1.X": picks the highest valid patch+gke.N patch in the 1.X version

When no valid version carries that prefix, there is nothing to pick, so the request should be rejected rather than stored.

## Actual behavior

`GkeService.resolveMasterVersion` keeps the alias verbatim when it does not match the advertised version:

```java
if (VERSION_PREFIX_ALIAS.matcher(requested).matches()) {
    return DEFAULT_MASTER_VERSION.startsWith(requested + ".")
            || DEFAULT_MASTER_VERSION.startsWith(requested + "-")
            ? DEFAULT_MASTER_VERSION : requested;
}
```

So `{"nodeVersion": "1.27"}` stores `1.27` as the pool's version, and the emulator then reports a version it does not consider valid: `getServerConfig()` advertises exactly one entry in both `validNodeVersions` and `validMasterVersions`, the default `1.30.5-gke.1014001`. A client that upgrades to `1.27` and then reads the pool back sees a version the same emulator would reject on validity grounds.

## Reproduction

```bash
curl -sX PUT ".../clusters/c/nodePools/default-pool" -H 'content-type: application/json' -d '{"nodeVersion": "1.27"}'
curl -s  ".../clusters/c/nodePools/default-pool" | python3 -c 'import json,sys; print(json.load(sys.stdin)["version"])'
# 1.27          <- an unmatched prefix alias, stored as if it were a version
curl -s ".../serverConfig" | python3 -c 'import json,sys; print(json.load(sys.stdin)["validNodeVersions"])'
# ['1.30.5-gke.1014001']
```

## Suggested direction

Reject an unmatched `1.X` prefix with 400 `INVALID_ARGUMENT`, in the same shape the resolver already uses for a malformed version, naming the field the client sent (that part landed in #199).

This is behaviour the current tests pin, so the change also updates `GkeServiceTest.updateNodePoolResolvesNodeVersionAliasesAgainstTheMaster` (the `"1.27"` case added in #230) and the equivalent assertions for `desiredNodeVersion` from #198. That is the main reason this is filed separately rather than folded into either PR.

**Unverified:** real GKE's exact error text and status for an unmatched prefix. The proto settles what the alias means but not what the service returns when nothing matches, and no capture exists under `local/google`. Worth a live capture before choosing the message.

## Environment

- Found while reviewing #230

avison9:
## Summary

The `1.X` / `1.X.Y` version alias means "the highest valid version under this prefix" (cluster_service.proto L4317-L4325). When no valid version carries the prefix there is nothing to pick, but `resolveMasterVersion` stored the alias as sent, so `{"nodeVersion": "1.27"}` left the pool reporting `1.27` while `getServerConfig()` advertised only `1.30.5-gke.1014001`. The unmatched alias is now rejected with 400 `INVALID_ARGUMENT`, on all four version fields, in the resolver's existing message shape, naming the field and the valid version.

Closes #231

```bash
# BEFORE
curl -sX PUT ".../clusters/c/nodePools/default-pool" -H 'content-type: application/json' -d '{"nodeVersion": "1.27"}'
curl -s  ".../clusters/c/nodePools/default-pool" | python3 -c 'import json,sys; print(json.load(sys.stdin)["version"])'
# 1.27          <- stored; validNodeVersions is ['1.30.5-gke.1014001']

# AFTER
# 400 {"error":{"code":400,"message":"Invalid nodeVersion \"1.27\": no valid version matches this alias (valid: 1.30.5-gke.1014001)","status":"INVALID_ARGUMENT"}}
```

**Root cause:** the resolver validated the alias grammar but not its meaning; `? DEFAULT_MASTER_VERSION : requested` fell back to storing the alias when the pick set was empty. That fallback was a deliberate choice in #197 and pinned by #198 and #230, which is why this is its own PR.

**What changed:**
- `GkeService.resolveMasterVersion`: the unmatched-prefix branch throws instead of returning the alias. Explicit `1.X.Y-gke.N` values outside the advertised one are still stored verbatim (a pin, not an alias, and what `createCluster` does).
- Tests: the `"1.3"` case in the UpdateMaster alias test and the two `"1.27"` lines from #230 flipped to rejection with the stored version asserted unchanged; new `unmatchedVersionAliasesAreRejectedOnEveryVersionField` (all four fields, message prefix, valid version named, nothing mutated, explicit pin still accepted); `"1.27"` added to the REST rejection loop.
- `docs/services/gke.md` updated.

On the open point in the issue: I had no GCP project to capture real GKE's exact wording against, so the status is the proto-settled 400 `INVALID_ARGUMENT` and the text follows the shape the resolver already uses. If a capture shows different wording, it is a one-string change.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

Reference: `cluster_service.proto` at googleapis `aa87617d67`, L4317-L4325 (`"1.X": picks the highest valid patch+gke.N patch in the 1.X version`). Before: an alias matching no valid version was stored and reported as a version. After: rejected with 400 `INVALID_ARGUMENT` on `masterVersion`, `desiredMasterVersion`, `desiredNodeVersion` and `nodeVersion`. Exact real-GKE error wording not captured (stated above). Verified through the Quarkus REST test against the real controller; the compat suites send explicit versions and `latest` only, so none changed.

## Checklist

- [x] `./mvnw test` passes locally: 1171/0/0/0 across 134 classes (1170 on `main` at adcec02 by derivation; the 1 new test is the only delta)
- [x] New or updated integration test added: `GkeUpdateNodePoolRestIntegrationTest` rejection loop extended with `"1.27"`; `GkeServiceTest.unmatchedVersionAliasesAreRejectedOnEveryVersionField`, plus three existing assertions flipped as the issue anticipated. The new test fails on `main` (`Expected GcpException to be thrown, but nothing was thrown`) and passes here.


avison9:
Right, and it is pre-existing on `main` rather than new here: `updateCluster` applies `desiredNodeVersion` to the pools before it resolves `desiredMasterVersion`, so any rejection on the master field, the resolver's existing bare-`1` rejection included, already leaves a half-applied update. This PR adds one more rejection path through that gap; it does not create it.

The ordering fix is what #236 (and the competing #235 on the same issue) does: read and resolve both fields, and the node pool targets, before the first setter. Whichever of those lands, this branch rebases on it and the combined case is covered by the tests there. I have not duplicated that reorder here to avoid a three-way conflict on the same lines; if you would rather this PR be independent, say so and I will add it.

hectorvent (approved):
Thanks for this, and for being straight about the part the reference does not settle.

The proto does settle the important half: it documents `1.X` as picking "the highest valid patch+gke.N patch in the 1.X version" on all four version fields, so when nothing valid carries the prefix there is genuinely nothing to return and a 400 is right. I also checked that a valid alias is unaffected, and that your tests un-pin the three places that had locked in the old accept-anything behaviour rather than adding new assertions around it, which is the harder and better thing to do.

One follow-up rather than a change here: `createCluster` still takes `initialClusterVersion` without resolving it, and the proto documents the same alias list for that field, so the same symptom survives there. Happy to file that separately.

No blockers from my side.

PR #250, opened 2026-09-21
feat(bigquery): serve the Storage Write API over gRPC

hectorvent:
## Summary

> **Stacked on #249**, which is stacked on #248, #247, #246 and #245. The base of this PR is `feat/bigquery-storage-read`, so the diff here is only the Storage Write change. Review the stack in order; I will rebase as each lands.

Implements `google.cloud.bigquery.storage.v1.BigQueryWrite` over gRPC: `CreateWriteStream`, `GetWriteStream`, `FinalizeWriteStream`, `BatchCommitWriteStreams`, `FlushRows` and the bidirectional `AppendRows`.

**All four stream types**, with the state machine each one implies: the `_default` stream (committed on append, no offsets), `COMMITTED`, `PENDING` (rows appear only after `BatchCommitWriteStreams`) and `BUFFERED` (rows appear at `FlushRows`). Offsets are tracked and mismatches return the typed `StorageError` the client expects rather than a bare status.

**Rows** arrive as `ProtoRows` with a `DescriptorProto`, decoded in `BigQueryProtoRows` against the destination schema. The decoder follows proto2 semantics, honors the `column_name` field option that lets clients send `col_0`-style placeholder names, and decodes the packed representations BigQuery uses: `CivilTimeEncoder` for DATETIME and TIME, and `BigDecimalByteStringEncoder`'s little-endian form for NUMERIC and BIGNUMERIC.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## GCP Compatibility

Grounded in `local/google/googleapis/google/cloud/bigquery/storage/v1/storage.proto` and `stream.proto` for the RPC set, `WriteStream.Type`, the `AppendRowsRequest`/`Response` shapes, offset semantics and the `StorageError` codes. The packed encodings were read from `local/google/google-cloud-java/java-bigquerystorage` (`CivilTimeEncoder`, `BigDecimalByteStringEncoder`), since the proto describes the field type but not the packing.

`BigQueryWriteGrpcIntegrationTest` (426 lines) drives each stream type through the **generated `BigQueryWriteGrpc` stub**, not the GAPIC client. The real `BigQueryWriteClient` and `JsonStreamWriter` are exercised by `compatibility-tests/sdk-test-java` (`BigQueryStorageTest`), which the native CI job runs.

**Native image:** the native binary failed at runtime with `Generated message class "StorageError" missing method "getCode"` while the JVM suite stayed green, so `StorageError` is explicitly `@RegisterForReflection`ed. `RowError` gets the same treatment: it is a nested field of `AppendRowsResponse` at the same depth, and no compat case exercises a rejected append natively, so nothing would catch it before a release build. The controller was also restructured so a response-building failure surfaces through `GcpGrpcController.grpcError` instead of leaving the client hanging, which is how the first bug showed up.

## Checklist

`./mvnw test`: 1287 tests, 0 failures, 0 errors, 0 skipped.

---

**Review round 2.** Fixes folded into this commit:

- **`BatchCommitWriteStreams` mutated stream state outside the per-stream lock.** Every other path guards `Stream` with its monitor; the commit path read and cleared `uncommitted` unguarded, so a concurrent append could be dropped, or the snapshot could throw `ConcurrentModificationException` and surface as `INTERNAL` rather than a `StorageError`. Streams are now locked in name order, so a multi-stream commit cannot deadlock against another.
- **`RowError` registered for native-image reflection**, per above.

**Corrections to this description.** Two claims were wrong:

1. It said `StorageError` "is only reachable through `google.protobuf.Any`". It is not: `storage.proto:656` declares `repeated StorageError stream_errors = 2;` on `BatchCommitWriteStreamsResponse`, a plain nested field of a top-level response. The registration was still needed, but the reason I gave for it was false, and that false reason is what scoped the fix to a single type. Re-checking under the correct theory is what surfaced `RowError`.
2. It said the in-repo test drives "a real `BigQueryWriteClient`". It drives the generated stub.

**Known and not fixed here:** write streams are never evicted, so an abandoned PENDING stream holds its rows for the process lifetime. The read side already sweeps by `expireTime`; the write side should mirror that, which I would rather do as its own change than bolt onto this one.


avison9 (changes requested):
@hectorvent
Thank you for this, and for the round two corrections to the description: tracing the `StorageError` registration back to the wrong theory and finding `RowError` from the right one is exactly how a native reflection gap should be closed. Driving every stream type through the generated stub, with the packed encodings built the way the client builds them, makes the suite read like the wire.

I checked the stream semantics against `storage.proto` rather than taking them on trust, and they hold: `_default` rejects finalize and flush, a lower offset is `ALREADY_EXISTS` and a higher one `OUT_OF_RANGE`, and `BatchCommitWriteStreams` commits nothing when any stream has an error ([storage.proto#L649-L656](https://github.com/googleapis/googleapis/blob/aa87617d67/google/cloud/bigquery/storage/v1/storage.proto#L649-L656)). I also read `CivilTimeEncoder` and `BigDecimalByteStringEncoder` in java-bigquerystorage: the shifts, masks, little endian byte order and scales 9 and 38 all match `BigQueryProtoRows`.

Greptile's "Batch commit races appends" was written against 0bbdbdf and is resolved by the ordered stream locks, so it can be skipped. "Stream state bypasses storage" I would not act on: read sessions keep their state the same way, and `docs/services/bigquery.md` states that uncommitted rows do not survive a restart. Its "Duplicate streams duplicate rows" still applies, below.

1. **(blocking)** `BigQueryStorageWrite.batchCommitWriteStreams` persists a stream once per time it is listed. I ran it on your branch: two rows appended to a PENDING stream, finalized, then committed with `write_streams` naming that stream twice, leaves four rows in the table and returns a `commit_time`. The same method also checks `committed` and `finalized` before `commit` takes the stream monitors, so two concurrent commits of one stream can both pass the check. I confirmed it with a probe that held the stream's monitor while two commits arrived on separate connections: both returned a `commit_time`. The second appends nothing, since `uncommitted` is already clear, but it still reports success where the proto says a stream "cannot be committed multiple times" ([storage.proto#L203](https://github.com/googleapis/googleapis/blob/aa87617d67/google/cloud/bigquery/storage/v1/storage.proto#L203)). Would you deduplicate the names (or reject a repeat) and move the state checks inside `commit`, once every monitor is held, with a test that commits the same stream twice in one request? A line in the PR body stating the invariant (each PENDING row reaches the table exactly once, and only through a successful commit) next to the lock order you already describe would make the concurrency notes complete.

2. **(follow-up, separate PR)** Agreed on write stream eviction mirroring the read side's `expireTime` sweep as its own change.

3. A few small ones, none blocking: `BigQueryProtoRows` uses inline `java.math.RoundingMode` and `java.time.DateTimeException` with no name clash, so those can be imports; and the genuine clashes (`com.google.rpc.Status` against `io.grpc.Status`, and the storage `TableFieldSchema` and `TableSchema` against the model ones) want the short comment AGENTS.md asks for.

Locally BigQuery* is 161/0 on 9af6e53, the new `BigQueryWriteGrpcIntegrationTest` cases all fail against the `feat/bigquery-storage-read` source (7 of 7 past `seed`), the packaged jar with the DuckDB engine (`FLOCI_GCP_SERVICES_BIGQUERY_MOCK=false`) passes `BigQueryStorageTest` 3/0, and CI is green across all 12 checks.

PR #253, opened 2026-09-21
docs: add Compute Engine to service navigation

vedit:
## Summary

Add Compute Engine to the MkDocs Services navigation and the Services Overview matrix. Both entries link to the service documentation supplied by #207.

Requested as a separate follow-up in https://github.com/floci-io/floci-gcp/pull/207#issuecomment-5751781000. This PR changes only two lines in two documentation files.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Breaking change (`feat!:`)
- [x] Docs / chore

## GCP Compatibility

No runtime or API changes. The table identifies the existing native Compute v1 REST root.

## Validation

`mkdocs build --strict` passed with `docs/services/compute.md` from #207 included in the validation checkout. Both generated navigation and overview links point to the Compute page.

This PR is draft and depends on #207: its target page is not on main yet. The dependency page is not duplicated in this diff.

## Checklist

- [x] Documentation build validated with its prerequisite
- [x] Commit follows Conventional Commits


hectorvent (approved):
Thanks for splitting this out so cleanly, vedit. Now that #207 has landed, I merged this onto current main and `mkdocs build --strict` passes, with both the nav entry and the overview row resolving to the Compute page. The endpoint cell matches the controller root in `ComputeController`.

**(follow-up, separate PR)** The home page's "Supported Services" table in `docs/index.md` still has no Compute Engine row. If you'd like, a one-line row there would round this out; happy for it to be here or in its own PR.

No blockers from my side. Since the dependency is merged, would you mark this ready for review?

